PR #12, opened 2022-03-03
Delete unwanted files from repo

omarelmalki:
Deleted local.properties from repo as it was added by mistake in the initial commit.
Deleted userStories.txt from the repo as it is no longed needed now that we have filled the product backlog.


ha2san (approved):
LGTM

PR #42, opened 2022-03-06
add home page update

pmcintyr:
update contains logo image asset and camera/registration buttons


omarelmalki (changes requested):
This looks really good !

Just a small request:
Since according to our figma 'prototype', we are going to have separate activity for login/sign-up that the user would access after the home page, the home page should not have email and password fields. It should only have the buttons that would get the user to the appropriate activity.

I would suggest deleting them from the main activity/home page. What do you think ?

omarelmalki (approved):
Looks good to me ! Thank you.

Mouniraki (approved):
LGTM !

PR #47, opened 2022-03-09
Add Login to App via Google Sign-In Integration

omarelmalki:
Added Google Sign-In Integration to allow users to login to the app with their Google Account.
Added Hilt to the project for Automatic Dependency Injection.
Tested SignInActivity with dependency injections using Espresso and Hilt.


hamzahass:
The code looks good to me, well done ! I would suggest to delete the printlns in the SignInActivity if they are not necessary.

hamzahass (approved):
The code looks good to me, well done !

pmcintyr:
The test coverage for the SignInAccountModule seems to not reach 100% indicating an issue with SignInActivityTest, getLastSignedInAccount(Context context) returns null if the user has never signed in before or has signed out / revoked access. To reach complete test coverage, I suggest that we test the case where we sign in once, exit the app without signing out, then signing in again.

omarelmalki:
> The test coverage for the SignInAccountModule seems to not reach 100% indicating an issue with SignInActivityTest, getLastSignedInAccount(Context context) returns null if the user has never signed in before or has signed out / revoked access. To reach complete test coverage, I suggest that we test the case where we sign in once, exit the app without signing out, then signing in again.

Hi Paul, this is a good idea. Unfortunately, the line of code in question gets the lastSignedInAccount from Google API and thus we need to be signed in with a real Google Account in order for it to return something other than null. I have already tested this locally by logging in on my Google Account during the test, exiting and coming back as you suggested.

As you can see in the code, I have used Hilt to inject mock accounts as the lastSignedInAccount in order to test the same scenario without relying on the Google API function. This is how I managed to reach 97% coverage when the majority of the code used Google API functions.

Feel free to propose any other solution that would allow us to test the code with real google accounts via CI.

pmcintyr (approved):
Thank you for your answer, because the testing of this method worked locally for you, I can now approve this pull request! Cirrus seems to be a bit tricky sometimes ;)

PR #48, opened 2022-03-09
Added a new activity that allows the user to scan qr Code

ha2san:
New Activity added that allow the user to report an object by scanning
the qr code or to enter the ID of the object
A test was added too for this feature, but we must find a way to test the camera.


pmcintyr (changes requested):
We can try and test the case where the camera crashes, aka the Camera initialization error in QrCodeScannerActivity, this might bring up the test coverage, here is an article on how to deny permissions (to try and test the error case): https://alexzh.com/ui-testing-of-android-runtime-permissions/

ha2san:
Ok thank you, but I have already read this article and tried this on my emulators but it didn't work for me.
It seems that once the permission is granted, it remains that way. I will try to read more documentation on this and get back to you.
Other than this, do you see any others problems ?

pmcintyr (approved):
Thanks for the details! I don't see anything else to fix and I saw that you resolved the conflicts with the main branch already, good job!

omarelmalki (changes requested):
Great work !

However, I believe there is an issue with the constraints of your EditText and Report Button. I tested them on my Pixel 2 emulator and they are not visible when I run the app.

I would suggest setting their position relative to the bottom and sides of the screen so that it remains consistent on most devices.

To test the Camera actions, I would suggest looking at Dependency Injections to test your code with a Mock Camera Implementation to mock the camera's result.
Try to look at this link for some ways to test Camera Intents with Espresso : https://proandroiddev.com/testing-camera-and-galley-intents-with-espresso-218eb9f59da9

ha2san:
Hi Omar, 
I think that I fixed the issue the constraints, can you look one more time and tell me if it works.
Also thanks for the link, i will have a look to it

omarelmalki (approved):
Perfect, the constraints problem seems to work.
I'll approve the changes since the coverage you have is reasonable considering the camera dependency issue. We'll look into how to improve it.

Amazing work again !

PR #51, opened 2022-03-09
Adding UI for Item List Activity

Mouniraki:
Implementation of issue #39
Contains all elements from the Figma, needs to be bound with actual functionality (backend)


hamzahass:
The code looks good to me ! You can try to go over this post, it allows to check whether the keyboard is shown and it could help with the coverage : https://stackoverflow.com/questions/33970956/test-if-soft-keyboard-is-visible-using-espresso.

app/build.gradle
```diff
@@ -47,13 +47,16 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.4.1'
     implementation 'com.google.android.material:material:1.5.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.3'
+    implementation 'com.google.android.material:material:1.5.0'
+    implementation 'androidx.test.uiautomator:uiautomator:2.2.0'
+//    implementation 'androidx.test.espresso:espresso-contrib:3.4.0'
```
omarelmalki, line 52: Here, you should replace 'implementation' by 'androidTestImplementation'

omarelmalki (changes requested):
Amazing work on the item menu !

From what we have discussed, you tried using 'androidx.test.espresso:espresso-contrib:3.4.0' to test the delete item functionality of your activity but it made the build fail so you deleted the related tests and removed the dependency.

I tried on my own machine and could make the build succeed by changing how you add the dependency in grade.build. (see related comment)

Thus, I suggest that you try this and if it works re-add the related tests.

Again, really impressive work here.

app/build.gradle
Mouniraki: Thank you for the feedback ! I tried what you proposed for the dependency, but it seems that on my machine it still keeps crashing all the tests for whatever reason (may be related to the fact that our architectures are different, but I doubt it) when I do the modification. I will try to re-import the project from scratch and let you know here :)

omarelmalki: Sure. Do you want me to push the working code on the branch to make it easier?

Mouniraki: Yes for sure ! I see that you already made some changes, I will take a look at them and come back to you

Mouniraki: I don't know why, but by pulling your changes (which are exactly what I did and didn't work), everything now works flawlessly, and I can finally write my tests on the RecyclerView !! Thank you so much !

omarelmalki: Perfect !

hamzahass (approved):
The code looks great to me, good job on increasing the code coverage !

PR #52, opened 2022-03-09
Feature/settings activity

DeepGreen1:
Adds the settings menu. Is currently a template of the settings activity.


ha2san (approved):
Hi, nice work !
Do we still need the println in the MainActivity.kt file ? Apart from that, the code seems very good to me

pmcintyr (approved):
LGTM nice conflict handling

PR #53, opened 2022-03-09
Feature/item user architecture

hamzahass:
Added the Item and User class structures that will be used for the entire app.


omarelmalki (approved):
Great work ! Looks good to me.

ha2san (approved):
Really nice work, 
everything looks good to me. It's nice to see that your tests cover 100% of your codes

PR #55, opened 2022-03-10
Integrate User-Item model with Sign In

omarelmalki:
Fixed unclosed activity scenario in a few tests.
Integrated User class with SignIn Account result to check for compatibility and catch problems early.
Precise data needed for a user of the app is yet to be determined and will be decided when we implement the database.


app/src/main/java/com/github/brugapp/brug/SignInActivity.kt
```diff
@@ -74,12 +77,19 @@ class SignInActivity : AppCompatActivity() {
         }
     }
 
+    // return new Brug User from SignInAccount
+    private fun createNewBrugUser(account: SignInAccount?): User? {
+        if (account?.displayName == null || account.email == null || account.idToken == null) return null
+        return User(account.displayName!!, account.displayName!!, account.email!!, account.idToken!!)
```
Mouniraki, line 83: Maybe consider either refactoring the User class to only take 1 name, or try to get the real first and last names of the returned user after signing in

app/src/main/java/com/github/brugapp/brug/SignInActivity.kt
```diff
@@ -92,6 +102,7 @@ class SignInActivity : AppCompatActivity() {
         signInClient.signOut()
```
Mouniraki, line 102: Make sure here that the user is successfully signed out before performing other actions

Mouniraki (approved):
Not a lot of changes (good thing !), feel free to check my comments for the User code for next sprint while refactoring the class
Otherwise, looks good to me !

app/src/main/java/com/github/brugapp/brug/SignInActivity.kt
omarelmalki: Hi, thanks for the feedback !
You are right and both options are easily possible and are already planned for future commits. The goal of the pull request was simply to ensure a temporary link between sign in user and brug user to catch mistakes early.
Once we have the database implemented, it will be easier to know exactly which unified signature to keep.

ha2san (approved):
the changes made look good to me, nothing to complain about.

PR #75, opened 2022-03-12
add FireBase setup

pmcintyr:
add FireBase setup & configuration to the Unlost android app

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -73,6 +73,12 @@ dependencies {
     // For instrumented tests.
     androidTestImplementation 'com.google.dagger:hilt-android-testing:2.38.1'
     kaptAndroidTest 'com.google.dagger:hilt-android-compiler:2.38.1'
+
+    // Import the Firebase BoM
+    implementation platform('com.google.firebase:firebase-bom:29.2.0')
+
+    //add dependency for Firebase SDK
+    implementation 'com.google.firebase:firebase-analytics-ktx'
 }
 
 tasks.withType(Test) {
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -111,4 +117,6 @@ task jacocoTestReport(type: JacocoReport, dependsOn: ['testDebugUnitTest', 'crea
 
 connectedCheck {
     finalizedBy jacocoTestReport
-}
\ No newline at end of file
+}
+
+apply plugin: 'com.google.gms.google-services'
```
omarelmalki, line 122: It would be great if this line is at the top of the file right after the other plugins.

omarelmalki (approved):
Approved. Thanks a lot for adding it early.
Feel free to check out my comment so we can keep the build.gradle file organized.

app/build.gradle
pmcintyr: done✅

ha2san (commented):
Hi I see that you added some dependencies, I don't see any bug in the code but what is Firebase and why do we need it ?

pmcintyr:
> Hi I see that you added some dependencies, I don't see any bug in the code but what is Firebase and why do we need it ?

Hi there! We are using Cloud Firestore which is a part of Firebase that we use to store and sync our app's data in a NoSQL database!

ha2san (approved):
> Hi there! We are using Cloud Firestore which is a part of Firebase that we use to store and sync our app's data in a NoSQL database!

Very interesting!! I don't see any problem with that, can't wait to see the progress of this stage

PR #76, opened 2022-03-14
add Firebase User Registration

pmcintyr:
create User/Item collections in Firestore & integrate Firestore registration in project


app/build.gradle
```diff
@@ -56,13 +59,15 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.4.1'
     implementation 'com.google.android.gms:play-services-auth:20.1.0'
     implementation "com.google.dagger:hilt-android:2.38.1"
+    implementation 'com.google.firebase:firebase-auth:19.2.0'
```
DeepGreen1, line 62: It would be interesting to see if we can use the latest version of the firebase-auth API

app/src/main/java/com/github/brugapp/brug/sign_in/SignInResultHandler.kt
```diff
@@ -15,6 +15,7 @@ abstract class SignInResultHandler {
 class SignInResultHandlerGoogle : SignInResultHandler() {
     private var account: GoogleSignInAccount? = null
     override fun handleSignInResult(result: ActivityResult): SignInAccount {
+        println("I am in correct handler")
```
DeepGreen1, line 18: You should remove those comments

app/src/main/java/com/github/brugapp/brug/RegisterUser.kt
```diff
@@ -0,0 +1,103 @@
...
+                    Log.w(TAG, "createUserWithEmail:failure", task.exception)
+                    Toast.makeText(this@RegisterUser, "Authentication failed.", Toast.LENGTH_SHORT).show()
+                    progressBar?.visibility = View.GONE //updateUI(null)
+                }
+            }
+    }
+
+    fun anyEmpty(): Boolean {
+        if (firstnametxt.isEmpty()) { firstName?.error = "Please enter first name"
+            firstName?.requestFocus() //return
```
DeepGreen1, line 84: Are the "return" comments necessary?

DeepGreen1 (changes requested):
Good chunk of work!
Did you try the "optimize the imports" tool in the IDE?

omarelmalki (commented):
Nice work !
Feel free to check out the few comments I left.
Also, optionally, since you added and configured Firebase and Firestore, it would be great if you could add a page in the wiki to explain to the developers in the team how to use it.

For further PRs, as a general rule, I would suggest separating concerns by not coding everything in the Activity class. There is a page in the wiki that explains the app architecture we will be using when developing the app, you can already check it out.
But do not worry about it for this PR as there will be an example of how to do it in my next PR and we will adapt the code next sprint.

app/src/main/java/com/github/brugapp/brug/RegisterUser.kt
pmcintyr: you are right, these comments have now been deleted

app/src/main/java/com/github/brugapp/brug/sign_in/SignInResultHandler.kt
pmcintyr: you are right, these comments have now been deleted

app/build.gradle
pmcintyr: yes, as Omar stated in his code review :  since we are using the BoM, we do not need to specify versions in Firebase library dependencies.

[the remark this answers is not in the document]
pmcintyr: I agree with you, this has been corrected now!

app/src/main/java/com/github/brugapp/brug/User.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.brugapp.brug
+
+class User(firstnametxt: String, lastNametxt: String, emailtxt: String, passwordtxt: String) {
```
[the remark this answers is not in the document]
pmcintyr: I took a closer look, we never used this User class! I was using it for informational purposes only, we do no longer need this so the class has been removed.

DeepGreen1 (approved):
Good work, thanks for the changes.

omarelmalki (approved):
Great job, thanks for the changes. Feel free to check out Hilt (see wiki) to mock Firebase authentication to test the rest of your code.

app/src/main/AndroidManifest.xml
```diff
@@ -1,56 +1,62 @@
-<?xml version="1.0" encoding="utf-8"?>
```
Mouniraki, line 1: I don't really understand what are the real changes here, it seems that the whole file has been changed according to Git but I don't see much real changes

app/src/main/java/com/github/brugapp/brug/RegisterUser.kt
```diff
@@ -0,0 +1,104 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.firestore.ktx.firestore
+import com.google.firebase.ktx.Firebase
+
+
+class RegisterUser : AppCompatActivity(), View.OnClickListener{
+
+    private val db = Firebase.firestore
+    private var progressBar: ProgressBar? = null
```
Mouniraki, line 20: Consider using lateinit variables instead of initializing the values to null

app/src/androidTest/java/com/github/brugapp/brug/RegisterUserTest.kt
```diff
@@ -0,0 +1,101 @@
...
+        )
+            .respondWith(
+                Instrumentation.ActivityResult(
+                    Activity.RESULT_OK,
+                    Intent()
+                )
+            )
+
+        // check if text appears correctly
+        Espresso.onView(ViewMatchers.withId(R.id.newAccount))
```
Mouniraki, line 79: Avoid calling the Espresso functions with the Class names to trim your code a bit, since you already imported them in your test class

Mouniraki (commented):
Nice job ! Check all my different comments on the code for suggestions, otherwise everything looks good to me

app/src/main/AndroidManifest.xml
pmcintyr: Sorry about that, I made some modifications that I decided to retract. The file is identical!

app/src/main/java/com/github/brugapp/brug/RegisterUser.kt
pmcintyr: good point, i have corrected this in the most recent commit

app/src/androidTest/java/com/github/brugapp/brug/RegisterUserTest.kt
pmcintyr: good idea, i have modified this in the most recent commit

PR #78, opened 2022-03-15
Add UI for Chat Menu Activity

Mouniraki:
Implementation of issue #65.

This PR involves the following changes :
- Added new ChatMenuActivity activity, with its UI components in the associated XML file
- Added new chat_entry_layout.xml, specifically for list entries related to ChatMenuActivity
- Added new CustomTopBar class to avoid code duplications between ChatMenuActivity and ItemsMenuActivity
- Added a "drag-to-reorder" feature on the lists, with their tests
- Implemented transitions with Bottom Navigation Bar in ChatMenuActivity and ItemsMenuActivity (i.e., clicking on the different menus trigger a transition to the requested Activity)
- Refactored activity_items_menu.xml, to remove useless parents
- Removed ItemsViewModel and ItemsCustomAdapter, replaced by more general ListViewModel and ListCustomAdapter classes 
- Moved list callback implementation from ItemsMenuActivity to a new separate ListCustomCallback class to avoid code duplications with the Chat Menu UI implementation
- Moved ViewHolder implementation from the old ItemsCustomAdapter to a new separate ListViewHolder class

PR #81, opened 2022-03-16
Add UI for item information

ha2san:
new activity created and integrated. The activity gives information about the clicked item.
For future sprint, a link must be created between a ListViewModel and a Item


ha2san:
PR #78 must be approved first, as this branch depends on it

PR #82, opened 2022-03-16
UI added for item information

ha2san:
new activity created and integrated. The activity gives information about the clicked item.
For future sprint, a link must be created between a ListViewModel and a Item


app/src/main/java/com/github/brugapp/brug/ItemInformationActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+        setTextView(R.id.item_last_location, extras, "lastLocation")
+        setTextView(R.id.item_owner, extras, "owner")
+        setTextView(R.id.item_date, extras, "addedOn")
+        setTextView(R.id.item_description, extras, "description")
+        val icon = findViewById<ImageView>(R.id.img)
+        val img: Int = extras?.get("image") as Int
+        icon.setImageResource(img)
+    }
+
+    @SuppressLint("SetTextI18n")
```
omarelmalki, line 30: Why is this @SuppressLint annotation necessary ? I suggest adding comments to explain why is it necessary in the code.

app/src/androidTest/java/com/github/brugapp/brug/ItemsMenuActivityTest.kt
```diff
@@ -56,8 +59,9 @@ class ItemsMenuActivityTest {
     /* THESE NEED A CONFLICTING ESPRESSO DEPENDENCY */
     @Test
     fun swipeLeftOnItemTriggersSnackBar() {
+        Intents.init()
```
omarelmalki, line 62: Do you plan on adding new tests that use Intents in the future ?
If yes, it would be great to put the Intents.init() and Intents.release() functions into a @Before  and @After function respectively, to avoid code duplication and conflicts. You can keep it in mind for next time or change it now as it is not a problem yet.

```suggestion
// (outside the class)
@Before
fun setUp() {
    Intents.init()
}
```

app/src/androidTest/java/com/github/brugapp/brug/ItemsMenuActivityTest.kt
```diff
@@ -66,7 +70,14 @@ class ItemsMenuActivityTest {
...
         )
-        snackBar.check(matches(withText("Item \"Phone\" is deleted")))
+
+        Intents.intended(
+            Matchers.allOf(
+                IntentMatchers.toPackage("com.github.brugapp.brug"),
+                IntentMatchers.hasComponent(ItemInformationActivity::class.java.name)
+            )
+        )
+        Intents.release()
```
omarelmalki, line 80:
```suggestion
// (outside the class)
@After
fun finishUp() {
    Intents.release()
}
```

omarelmalki (commented):
Great work ! Feel free to check out my comments to improve documentation and tests.

ha2san:
thanks both of you, 
@DeepGreen1 I added some comments.
@omarelmalki I removed the code that was not necessary, and I keep in mind the remark you made to me for the next few times

app/src/androidTest/java/com/github/brugapp/brug/ItemInformationActivityTest.kt
```diff
@@ -0,0 +1,38 @@
...
+        ActivityScenario.launch<ItemInformationActivity>(intent).use {
+            onView(ViewMatchers.withId(R.id.item_name)).check(matches(withText("Phone")))
+            onView(ViewMatchers.withId(R.id.item_description)).check(matches(withText("Samsung Galaxy S22")))
+            onView(ViewMatchers.withId(R.id.tv_name)).check(matches(withText("Phone")))
+            onView(ViewMatchers.withId(R.id.item_last_location)).check(matches(withText(str)))
+            onView(ViewMatchers.withId(R.id.item_owner)).check(matches(withText(str)))
+            onView(ViewMatchers.withId(R.id.item_date)).check(matches(withText(str)))
+        }
+    }
+}
```
hamzahass, line 38: For the ItemInformationActivityTest, I would suggest to separate the checks into different Tests for future use, in case one of the fields fails the test it would be easier to point it out.

app/src/main/res/values/colors.xml
```diff
@@ -7,4 +7,19 @@
...
+    <color name="followers">#0D5D98</color>
+    <color name="headings">#3797CF</color>
+    <color name="solidCircle">#AAFFFFFF</color>
+    <color name="circleBorder">#50FFFFFF</color>
+    <color name="gradEnd">#67EFC8</color>
+    <color name="gradStart">#63C9FF</color>
+    <color name="followersBg">#65DCE5</color>
+    <color name="followingBg">#49BEFF</color>
+    <color name="grey">#E5E5E5</color>
+
```
hamzahass, line 24: Cirrus CI seems to say that some colors added on the colors.xml file are not actually used, in this case it would be good to delete these lines.

hamzahass (commented):
Good work ! I have listed 2 comments below, please go over them and follow up on the concerns (ps : for the second comment on colors.xml file there is the same remark for the strings.xml file so please also check this file)

app/src/androidTest/java/com/github/brugapp/brug/ItemInformationActivityTest.kt
ha2san: Thank you for your comment, I changed that as you requested in my last commit [fd6c7e1]

app/src/main/res/values/colors.xml
ha2san: We may need these colors in the future, so it might be a good idea to keep them

PR #84, opened 2022-03-17
Feature/chat activity

DeepGreen1:
- Added a new activity for the chat
- Added a XML file for the activity
- Added tests for the activity and changes in the build.gradle:app (in order to exclude protobuf-lite)
- Added a ChatItemModel class to represent the messages
- Added a XML file for the messages displayed in the activity
- Added a ChatItemAdapter (and ViewHolder) class to link data of the database to elements of the View


app/src/main/java/com/github/brugapp/brug/ChatActivity.kt
```diff
@@ -0,0 +1,140 @@
...
+        recyclerView.layoutManager = linearManager
+        recyclerView.setHasFixedSize(true)
+
+        chatArrayList = arrayListOf()
+        adapter = ChatItemAdapter(chatArrayList)
+
+        recyclerView.adapter = adapter
+
+        println("UNTIL HERE OK")
+        eventChangeListener()
```
Mouniraki, line 46: Consider putting this code in a method, and only do a call to the method in onCreate()
Also : make sure to do everything graphical in the XML as much as possible (here everything should be doable in the XML since it requires no interaction/callback definitions)

app/src/main/java/com/github/brugapp/brug/ChatItemAdapter.kt
```diff
@@ -0,0 +1,34 @@
+package com.github.brugapp.brug
+
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+
+// Adapter that binds the list of messages to the instances of ChatItemModel
+class ChatItemAdapter(private val messageList: ArrayList<ChatItemModel>) :
```
Mouniraki, line 10: Consider maybe renaming the class to something more in line with its real purpose, like for example "ChatMessagesListAdapter"

app/src/main/java/com/github/brugapp/brug/ChatItemModel.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.brugapp.brug
+
+data class ChatItemModel(
+    var content: String? = null,
+    var datetime: String? = null,
+    var sender: String? = null
+)
```
Mouniraki, line 7: Declaring the arguments as "var" is useless here, unless you assumed that the sender and datetime can be changed maybe (doesn't really make sense in my opinion) ?

app/src/main/res/layout/activity_chat.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Mouniraki, line 1: Be careful with the String warnings, you should absolutely create a new entry in the strings.xml file instead of hard-coding the strings in the actual layout of your Activity (simple alt-click to trigger the suggestion in Android Studio)

app/src/main/res/layout/chat_item_layout.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Mouniraki, line 1: Same as for the Adapter class, try to find better names for the classes/layout files

Mouniraki (changes requested):
Nice job ! Check my comments and try to do the mentioned changes, there's only a few steps before approval :)

app/src/main/java/com/github/brugapp/brug/ChatActivity.kt
DeepGreen1: I had to do it without the graphical XML so set the stackFromEnd. Otherwise, I agree with you and will create a method, thanks!

app/src/main/res/layout/activity_chat.xml
DeepGreen1: Agreed, I forgot to change this

app/src/main/java/com/github/brugapp/brug/ChatItemAdapter.kt
DeepGreen1: Good idea!

app/src/main/java/com/github/brugapp/brug/ChatItemModel.kt
DeepGreen1: Good point :)

app/src/main/java/com/github/brugapp/brug/ChatItemModel.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.brugapp.brug
+
+data class ChatItemModel(
+    val content: String? = null,
+    val datetime: String? = null,
+    val sender: String? = null
+)
```
Mouniraki, line 7: Setting the values to null and having nullable values here doesn't make sense, you can't have an empty message in the conversation and every message has a timestamp & a sender. Remove the "= null" assignment and remove the "?" after each parameter type

Mouniraki (changes requested):
Almost there, perform the last changes and the PR will be ready to be merged :)

app/src/main/java/com/github/brugapp/brug/ChatMessagesListAdapter.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun getItemCount(): Int {
+        return messageList.size
+    }
+
+    public class ViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+        val sender: TextView = itemView.findViewById(R.id.chat_item_sender)
+        val datetime: TextView = itemView.findViewById(R.id.chat_item_datetime)
+        val content: TextView = itemView.findViewById(R.id.chat_item_content)
+    }
```
hamzahass, line 33: For the class ViewHolder, why did you create it inside the ChatMessagesListAdapter and not as a class itself ?

app/src/main/java/com/github/brugapp/brug/ChatItemModel.kt
DeepGreen1: Actually if you do this, the activity will crash at creation time. Also, I'll refarctor this class when we will link the ChatMenuActivity, the ChatActivity and the authentification system: currently, there is no check if the message/sender is an empty string; it'll be more coherent to do everything together.

hamzahass (commented):
This is good work ! I left out comments and didn't request changes as for some parts I was not sure what was the reasoning behind your decisions, please explain this and follow up on my comments.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    fun checkIfMessageFieldIsPresent() {
+        onView(withId(R.id.editMessage)).check(matches(withId(R.id.editMessage)))
+        onView(withId(R.id.editMessage)).check(matches(withHint("Message")))
+    }
+
+    @Test
+    fun checkIfSendButtonIsPresent() {
+        onView(withId(R.id.buttonSendMessage)).check(matches(withId(R.id.buttonSendMessage)))
+        onView(withId(R.id.buttonSendMessage)).check(matches(withText("Send")))
```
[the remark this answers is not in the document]
DeepGreen1: I was testing if the button was actually there. However you're right, the next would also fail the test if it's not present.

Mouniraki (approved):
Based on your last comment I approve this PR for merging, we will make sure that one member of the team will be in charge of refactoring the Chat part to solve the little issues in the following sprints :)

app/src/main/java/com/github/brugapp/brug/ChatMessagesListAdapter.kt
DeepGreen1: As the ChatMenuActivity would use the same pattern, I wanted to make sure that there is no conflict. As stated in the PR 78, it would be great to see if we can refactor some classes in order to make better and more maintainable code, eg. in the next sprint.

app/src/main/java/com/github/brugapp/brug/ChatActivity.kt
```diff
@@ -0,0 +1,142 @@
...
+            .addOnFailureListener { e ->
+                Log.w(TAG, "Error adding document", e)
+            }
+    }
+
+    private fun eventChangeListener() {
+        db = FirebaseFirestore.getInstance()
+        // TODO: Change the document when ChatListActivity is implemented
+        db.collection("Chat").document("User1User2").collection("Messages")
+            .orderBy("datetime", Query.Direction.ASCENDING)
```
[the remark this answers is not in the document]
DeepGreen1: Good question! Exactly, when we will be able to use the authentification part, we would create a document for each pair of users in which we will store the messages in the "Messages" collection.

To use a hash of the ID of the 2 users (performing the hash of the ID that are sorted) is practical so that we can always retrieve the messages as long as we know the ID of the participants of a conversation.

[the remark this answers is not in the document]
hamzahass: Okay that sounds good, thanks.

app/src/main/java/com/github/brugapp/brug/ChatMessagesListAdapter.kt
hamzahass: Okay it makes sense, thanks.

pmcintyr (commented):
Good job!

pmcintyr (approved):
Hey, awesome work with the chat!

For **ChatActivity.kt**, as Hamza also mentioned we will need to fix the hardcoded "_User1User2_" firestore document title to the respective chat documents in a future sprint.
I would suggest using the concatenated userID strings that we create when we create a new account.

I had trouble understanding the following code, could you explain to me what this code snippet from **ChatActivityTest** does?
```
        // Check the message
        onData(
            allOf(
                `is`(instanceOf(Map::class.java)), hasEntry(
                    equalTo("STR"),
                    `is`("Sender: TestSender")
                )
            )
        )
```

DeepGreen1:
For the last comment: this test will retrieve data from firebase and create a Map using the collected data. The test is then to check if one of the values of the map is "Sender: TestSender" (which is the case as we sent a message before this).

Remark for later: it would be great to see if a refactor is possible as said in the PR 78; also to update how messages are retrieved from firebase (see the last comment of Paul).

PR #104, opened 2022-03-18
Refactor ChatActivity (Separation of UI/ViewModel/Data)

DeepGreen1:


hamzahass (commented):
From what I have understood from reading the guide to the app architecture, the UI layer should contain an Activity layer (that should contain Activities and xml files) and a the View Model layer. Hence I don't understand why : First the ViewModel Layer is not a sublayer of the UI Layer, and why there is no Activity Layer (as a sublayer of the UI layer) that contains both Activity classes and the corresponding xml files ?

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+        val linearManager: LinearLayoutManager = LinearLayoutManager(this)
+        linearManager.stackFromEnd = true
+        recyclerView.layoutManager = linearManager
+        recyclerView.setHasFixedSize(true)
+
+        recyclerView.adapter = viewModel.getAdapter()
+    }
+
+    fun updateData(numberOfMessages: Int) {
+        print("THIS IS DONE")
```
omarelmalki, line 55: I would suggest removing this debug message.
```suggestion
```

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+        // Create a new message
+        val message = hashMapOf(
+            "sender" to sender,
+            "content" to content,
+            "datetime" to datetime
+        )
+
+        // Add a new document i.e. message
+        db = FirebaseFirestore.getInstance()
+        // Firebase.firestore change for kotlin?
```
omarelmalki, line 83: Here I would suggest using the kotlin version of firebase/firestore attributes and methods, like you have commented.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.github.brugapp.brug.R
+import com.github.brugapp.brug.model.ChatMessage
+import com.github.brugapp.brug.ui.ChatActivity
+import com.google.firebase.firestore.*
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class ChatViewModel : ViewModel() {
+
+    // TODO: Have to be removed when the data.DataBase class is implemented
```
omarelmalki, line 20: I see you have many TODO: comments for future changes, I would suggest also adding tasks in the scrum board so we do not forget about them.

omarelmalki (commented):
Hi, Great work on the refactoring !
First, since we still do not have a repository to get data from the database, we cannot do it yet, but do not forget to change all Firestore related methods by the abstract methods of the repository once it is implemented.

Second, I would suggest moving the ChatMessagesListAdapter.kt file to the 'model' package that contains the abstractions we use in the app, to only keep ViewModel classes inside the view_model package.

Finally, feel free to check the few comments I left in the code even though they are not directly related to architecture.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+    private lateinit var db: FirebaseFirestore
+
+    private lateinit var chatArrayList: ArrayList<ChatMessage>
+    private lateinit var adapter: ChatMessagesListAdapter
+
+    public fun getAdapter(): ChatMessagesListAdapter {
+        chatArrayList = arrayListOf()
+        adapter = ChatMessagesListAdapter(chatArrayList)
```
Mouniraki, line 28: The getAdapter() in my opinion should only be a getter to the Adapter, and it shouldn't also do any initialization. Consider having another method only doing the initialization of the list and Adapter (like in a constructor for instance) to have a better split of operations in each method.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+                    // TODO: Would be cleaner to have this update inside of ChatActivity
+                    //activity.updateData(adapter.itemCount - 1)
+                    // Instead I have to do this
+                    val rv = activity.findViewById<View>(R.id.recyclerView) as RecyclerView
+                    rv.smoothScrollToPosition(adapter.itemCount - 1)
+                }
+            })
+    }
+
+    @SuppressLint("CutPasteId")
```
Mouniraki, line 66: I have seen many times this statement in your code, I think you should maybe address the Lint issues instead of putting this statement to remove them, or at least justify why you have to put it by adding a little comment.

Mouniraki (changes requested):
Overall, great refactoring job. I agree with Omar's comments also, feel free to check my comments for more suggestions.
Also : don't forget to assign yourself to the PR !

DeepGreen1:
> From what I have understood from reading the guide to the app architecture, the UI layer should contain an Activity layer (that should contain Activities and xml files) and a the View Model layer. Hence I don't understand why : First the ViewModel Layer is not a sublayer of the UI Layer, and why there is no Activity Layer (as a sublayer of the UI layer) that contains both Activity classes and the corresponding xml files ?

The idea that is presented in the wiki is to separate the models / UI / ViewModels / Data modules
Thus the refactoring is about splitting my ChatActivity in different classes: ChatActvity and his corresponding XML that only concerns UI; ChatViewModel that handle the logic for the View = Activity (ViewModel package); ChatMessage and ChatMessageListAdapter that are abstractions used in the application (Model package)

I still need to update the viewModel to use the data package instead of calling firebase directly, but it's still not implemented.
Hopes this helps to understand!

DeepGreen1:
> Hi, Great work on the refactoring ! First, since we still do not have a repository to get data from the database, we cannot do it yet, but do not forget to change all Firestore related methods by the abstract methods of the repository once it is implemented.
> 
> Second, I would suggest moving the ChatMessagesListAdapter.kt file to the 'model' package that contains the abstractions we use in the app, to only keep ViewModel classes inside the view_model package.
> 
> Finally, feel free to check the few comments I left in the code for future work, even though they are not directly related to architecture.

Thanks for your comment! For your first comment, I already put some comments about this in the code, no worries! Concearning the second, that's a good idea. I'll update this in the next commit.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
DeepGreen1: Correct! I forgot to remove this, thanks!

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: Good comment

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: It's done, thanks

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: It's a good idea indeed! I changed this, thanks

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: This annotation is present each time that I'm getting text from EditText elements because apparently, errors could come due to copy/paste actions. They can be ignored (and that's what the annotation does).

omarelmalki (approved):
Perfect, thanks for the changes !

Mouniraki (approved):
Great job, we now have a great starting point regarding the code organization for the Chat part. PR approved for merging :)

hamzahass (approved):
Thanks for answering the question, much clearer now.

PR #109, opened 2022-03-20
Refactor Items Menu UI & Chat Menu UI following App Architecture

Mouniraki:
Addressing issue #98 partially.

- Moved ItemsMenuActivity & ChatMenuActivity to the ui package
- Added a ItemsViewModel class in the view_model package, handling the items list initialization & all the list callback actions specific to the ItemsMenuActivity class
- Added a ChatViewModel class in the view_model package, handling the chat list initialization & all the list callback actions specific to the ChatMenuActivity class
- Added a Conversation mock class in the model package to model a conversation between two users, its total completion will be addressed in another PR
- Splitted the original ListCustomAdapter class into two more specific ItemsListAdapter & ChatListAdapter classes located in the view_model package, to better address how an Item (resp. a Chat Conversation) is displayed in the list
- Added a new package ui.components containing the CustomTopBar class
- Added a new BottomNavBar class in the ui.components package to reduce code duplication
- Removed ListViewModel, now using the proper Item and ChatMessage classes to display elements in the lists
- Refactored ChatMenuActivityTest, to remove some tests related to dragging (feature removed for the moment only in ChatMenuActivity but still present in ItemsMenuActivity, the usefulness of it isn't really justified for the moment for conversations)


app/src/main/java/com/github/brugapp/brug/view_model/ChatListAdapter.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ListViewHolder {
+        val view = LayoutInflater.from(parent.context)
+            .inflate(R.layout.chat_entry_layout, parent, false)
+        return ListViewHolder(view, R.id.chat_entry_profilepic, R.id.chat_entry_title, R.id.chat_entry_desc) {
+            onItemClicked(chatList[it])
+        }
+    }
+
+    // Binds the list items to a view
+    @SuppressLint("SetTextI18n")
```
ha2san, line 28: Like @omarelmalki told me once, it would be better to explain what this line means and why do we need it

ha2san (approved):
really nice work, I find it much easier to understand the code and work with it. I just had one comment but you can add the change while merging with the name

app/src/main/java/com/github/brugapp/brug/view_model/ChatMenuViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+            Snackbar.make(viewHolder.itemView, checkText, Snackbar.LENGTH_LONG)
+                .setAction("Undo") {
+                    list.add(position, deletedElt)
+                    listViewAdapter.notifyItemInserted(position)
+                }.show()
+
+        }
+
+        // Drawing the red "delete" background when swiping to delete
+        override fun onChildDraw(
```
omarelmalki, line 92: Hi, since drawing the background can be considered as rendering elements on the UI, is it possible to move this functionality to the 'Activity' layer.

omarelmalki (commented):
Hi, Great Job on the Refactoring !
Even though it is not always 100% possible, try to keep the UI rendering elements in the "Activity" layer (Activity/.xml files) from the UI logic in the ViewModel.
See example in the comment below.

pmcintyr (approved):
Overall I agree with the changes you made in this PR.
There are a few conflicts with the main to handle.
Here are a couple notes I made:

**Conversation.kt**

If this refactor already brings the better model of conversation, the way I understand this, we can now get rid of the TODO.
If we are talking about a different refactor, we can leave the TODO comment.

`//TODO: Refactor the class to have a better model of a conversation`



**ChatMenuViewModel.kt**

What is meant when we say we move in a color database?
Does swipeBG make colour animations/transitions?

`private val swipeBG: ColorDrawable = ColorDrawable(Color.parseColor("#15b800")) // To move in a color database (maybe ?)`

Mouniraki:
> Overall I agree with the changes you made in this PR. There are a few conflicts with the main to handle. Here are a couple notes I made:
> 
> **Conversation.kt**
> 
> If this refactor already brings the better model of conversation, the way I understand this, we can now get rid of the TODO. If we are talking about a different refactor, we can leave the TODO comment.
> 
> `//TODO: Refactor the class to have a better model of a conversation`
> 
> **ChatMenuViewModel.kt**
> 
> What is meant when we say we move in a color database? Does swipeBG make colour animations/transitions?
> 
> `private val swipeBG: ColorDrawable = ColorDrawable(Color.parseColor("#15b800")) // To move in a color database (maybe ?)`

Thank you for your feedback. For the **Conversation** class the model isn't totally complete and I plan on completing it for my second task of the sprint. The second point related to the **ChatMenuViewModel** will be addressed in the future (and hopefully last) commit.

app/src/main/java/com/github/brugapp/brug/view_model/ChatListAdapter.kt
Mouniraki: Thank you for your comment, I addressed this issue in my last commit.

PR #110, opened 2022-03-21
Refactor registration activity to UI/ViewModel/Model layers

pmcintyr:
refactor the registerUser class into UI/ViewModel/Model as seen in the App's architecture in issue https://github.com/BrugApp/brug/issues/98.


app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_register_user)
+        val regUser = findViewById<Button>(R.id.registerbutton)
+        regUser.setOnClickListener(this)
+        progressBar = findViewById(R.id.progressBar)
+    }
+
+    override fun onClick(v: View?) { //we clicked the register button
+        viewModel.storeUserInput(
+            FirebaseAuth.getInstance(),
```
DeepGreen1, line 34: FirebaseAuth.getInstance() is java code, you could update it for kotlin code: Firebase.auth (see here: https://firebase.google.com/docs/auth/android/start#kotlin+ktx_2 )

Also, be careful with the next line (35) that I think, would return the same database (kotlin code this time)

app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+        if (!viewModel.anyEmpty()) {
+            onClickHelper()
+        }
+    }
+
+    private fun onClickHelper() {
+        progressBar.visibility = View.VISIBLE
+        //make an authentication account with email password tuple
+        viewModel.createAuthAccount()
+            .addOnCompleteListener(this) { task ->
```
DeepGreen1, line 50: To have even better separation of UI/Logic, you could try to split this method. If it's possible, the createAuthAccount() should handle all the logic that is made here: call addRegisterUserTask and print log text where the onClickHelper should only update the progressBar, show the toast and call the method inside the viewmodel.

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+import android.widget.ProgressBar
+import androidx.lifecycle.ViewModel
+import com.google.android.gms.tasks.Task
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.firestore.DocumentReference
+import com.google.firebase.firestore.FirebaseFirestore
+
+class RegisterUserViewModel : ViewModel() {
+    private lateinit var db: FirebaseFirestore
+    private lateinit var progressBar: ProgressBar
```
DeepGreen1, line 13: To separate view/Logic, the viewmodel should not contain any UI element (progressBar / EditText fields). You can send these values as parameter when you call the method from the activity.

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+                email.requestFocus()
+            }
+            else -> {
+                return false
+            }
+        }
+        return true
+    }
+
+    fun addRegisterUserTask(userToAdd: HashMap<String, Any>): Task<DocumentReference> {
```
DeepGreen1, line 93: Maybe add a TODO and/or a task in the scrum board to think to update this part of the code when the database class is created. (db will be changed from FirebaseFirestore to our new fancy class :) )

DeepGreen1 (dismissed):
Good work for the refactoring!
I let you some comments to read; l can approve the request if we agree to do them with the next refactor (the one due to the new "database" class), as we are all depending on the architecture changes before implementing the actual weekly work.

ha2san (approved):
Really nice work, I have nothing to add to what @DeepGreen1 said. I'm not an expert in app architecture but it seems adequate to me.

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
pmcintyr: This causes lots of **code duplication** when passing the same parameters to multiple functions but i did it anyway. If this issue comes up, I will redirect them here.

app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
pmcintyr: done

app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
pmcintyr: done in https://github.com/BrugApp/brug/pull/110/commits/d326ba81c162987a9633c1da83820ccc7995397b

app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+        setContentView(R.layout.activity_register_user)
+        val regUser = findViewById<Button>(R.id.registerbutton)
+        regUser.setOnClickListener(this)
+        progressBar = findViewById(R.id.progressBar)
+    }
+
+    override fun onClick(v: View?) { //we clicked the register button
+        viewModel.storeUserInput(
+            Firebase.auth,
+            Firebase.firestore,
```
omarelmalki, line 36: Here, the Activity layer (Activity class + xml file) should not be aware that the viewModel uses Firebase. I suggest keeping Firebase.auth and Firebase.firestore as attributes in the RegisterUserViewModel class so you do not have to put them as argument.
```suggestion
```

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -0,0 +1,123 @@
...
+import android.widget.Toast
+import androidx.lifecycle.ViewModel
+import com.google.android.gms.tasks.Task
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.firestore.DocumentReference
+import com.google.firebase.firestore.FirebaseFirestore
+
+class RegisterUserViewModel : ViewModel() {
+    private lateinit var db: FirebaseFirestore
+    private lateinit var mAuth: FirebaseAuth
```
omarelmalki, line 17:
```suggestion
    private val db: FirebaseFirestore = Firebase.firestore
    private val mAuth: FirebaseAuth = Firebase.auth
```

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -0,0 +1,123 @@
...
+            "firstName" to firstnametxt,
+            "lastName" to lastnametxt
+        )
+        return userToAdd
+    }
+
+    //stores user input data in the view model
+    fun storeUserInput(
+        mAuth: FirebaseAuth,
+        db: FirebaseFirestore,
```
omarelmalki, line 40:
```suggestion
```

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -0,0 +1,123 @@
...
+        firstName: EditText,
+        lastName: EditText,
+        password: EditText,
+        email: EditText
+    ) {
+        this.db = db
+        this.mAuth = mAuth
```
omarelmalki, line 47:
```suggestion
```

omarelmalki (changes requested):
Great job on the refactoring !
Make sure to preserve the abstraction between each layer. Checkout my suggestions to make sure the activity does not know the rest of the app is using Firebase.

app/src/main/java/com/github/brugapp/brug/ui/RegisterUserActivity.kt
pmcintyr: good idea! thank you for this feedback, i have now implemented it in this commit: https://github.com/BrugApp/brug/pull/110/commits/4188a218e6f70b8aae558fc29f648dce21812b74

PR #112, opened 2022-03-23
add firebase interface & solve conflicts

pmcintyr:
change firebase functions to call firebasehelper


app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
```diff
@@ -7,29 +7,22 @@ import android.widget.EditText
...
+import com.github.brugapp.brug.data.FirebaseHelper
 
 class RegisterUserViewModel : ViewModel() {
-    private val db: FirebaseFirestore = Firebase.firestore
-    private val mAuth: FirebaseAuth = Firebase.auth
+    private val helper = FirebaseHelper()
     private var firstnametxt = ""
     private var lastnametxt = ""
     private var emailtxt = ""
     private var passwordtxt = ""
```
Mouniraki, line 17: Try maybe to declare those variables as `private lateinit var` instead of assigning empty values to them if you are initializing them in a method/constructor afterwards

Mouniraki (changes requested):
Great job ! This will make the interfacing with Firebase much easier, but consider the comment I added in your code before I approve your PR :)

hamzahass (changes requested):
Good job, this firebase helper will be very useful to all of us! However I have two issues : 
- The first one is regarding the functions of the helper class, I think it would be good to have comments to specify what functions do specifically, as not all of us worked on the database.
- Secondly, for JUnit tests I would suggest to use assertThat with the Is matcher, rather tham assert(... == ...), what do you think ?

Mouniraki (approved):
Perfect, the PR is approved 👍

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    //returns Firestore authentication
+    fun getFirebaseAuth(): FirebaseAuth {
+        return mAuth
+    }
+
+    //returns Firestore database
+    fun getFirestore(): FirebaseFirestore {
+        return db
+    }
```
omarelmalki, line 26: This function should not be available. The goal of the Repository is to get the data from Firestore without giving direct access to it. If a component needs to use the database to get a User or an Item, it should call this a method in this class instead.

omarelmalki (changes requested):
Great job on the interface. Make sure you do not break the abstraction by giving direct access to the database when it should be an interface for it.

hamzahass (approved):
Thanks for adding comments and modifying the tests as requested, well done !

app/src/main/java/com/github/brugapp/brug/view_model/RegisterUserViewModel.kt
pmcintyr: Sounds like a good idea, i have now added this in a later commit!

pmcintyr:
> Good job, this firebase helper will be very useful to all of us! However I have two issues :
> 
> * The first one is regarding the functions of the helper class, I think it would be good to have comments to specify what functions do specifically, as not all of us worked on the database.
> * Secondly, for JUnit tests I would suggest to use assertThat with the Is matcher, rather tham assert(... == ...), what do you think ?

I agree with these suggestions and have implemented them in a more recent commit!

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
pmcintyr: I have deleted this function in the latest commit & have removed the corresponding tests.

omarelmalki (approved):
Thanks for the changes.

PR #114, opened 2022-03-24
Link Conversations in ChatMenuActivity to ChatActivity

Mouniraki:
Addresses issue #95.

This PR involves the following changes:
- Refactored ChatMenuActivity to send a Conversation object when a user clicks on a conversation
- Refactor User, Item, ChatMessage & Conversation to make them inherit the Serializable interface
- Refactored ChatViewModel, use of FireBase is on stand-by for the moment
- Refactored ChatActivity, now displaying the name of user & related item for each conversation in ActionBar
- Refactored activity_chat.xml to remove the Sender textbox
- Refactored constants used in multiple classes to avoid values duplication
- Refactored ChatMenuActivityTest & ChatActivityTest to add/modify tests for the new behaviors
- Add line in build.gradle to be able to use LocalDateTime methods on older APIs


app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -1,59 +1,151 @@
...
+                    ?: // has no item on such position
+                    return false
+                return itemMatcher.matches(viewHolder.itemView)
+            }
+        }
     }
 
+
+
+    /* OLD TESTS USING FIREBASE */
```
ha2san, line 113: It would be better to say why we don't use the old tests no more

app/src/main/java/com/github/brugapp/brug/model/Item.kt
```diff
@@ -1,9 +1,10 @@
 package com.github.brugapp.brug.model
 
+import java.io.Serializable
 import java.lang.IllegalArgumentException
 
 class Item// Generate the id later
-    (private var name: String, private var id: Int, private var description : String) {
+    (private var name: String, private var id: Int, private var description : String): Serializable {
```
ha2san, line 7: Why do theses classes need to be serializable ?

ha2san (approved):
Awesome work, I just have one question.It would be nice if you could explain it to me, I let a comment for that. I let another comment where I found that the code lacks some documentation in one place. Other than that, everything looks great to me.

app/src/main/java/com/github/brugapp/brug/model/Item.kt
Mouniraki: The classes need to be serializable to be able to pass them inside Intents. Initially I wanted to "implement" the Serializable interface only to Conversation class, but I had to do it also to every Conversation's attribute type (namely Item, User and ChatMessage)

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
Mouniraki: You are right, I will explain a bit more in the code why, but it is simply because this PR overrides all code related to Firebase which was written by @DeepGreen1 while keeping his implementation for a future refactoring involving links with the final Firebase database.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -1,25 +1,33 @@
...
+    fun sendMessage(content: String){
+        // TODO: Change the sender text to something related to the actual user in the future
+        val newMessage = ChatMessage(content, LocalDateTime.now(), "Me")
+        messages.add(newMessage)
+        adapter.notifyItemInserted(messages.size-1)
+    }
+
+
+
+/* OLD IMPLEMENTATION USING FIREBASE -> TO BE REFACTORED PROPERLY AFTERWARDS
```
omarelmalki, line 30: Again, why not keep the functionality uncommented and refactor afterwards ?

app/src/main/java/com/github/brugapp/brug/model/Item.kt
```diff
@@ -1,9 +1,10 @@
 package com.github.brugapp.brug.model
 
+import java.io.Serializable
```
omarelmalki, line 3: I assume you made Items Serializable to be able to send them as Extras between intents, right ? But why do we need to send them between intents? Can't we get them from a public method in the reporsitory/database every time we need them ?

omarelmalki (commented):
Great job on the linking. I have a few questions, please check them below in my comments.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -1,59 +1,151 @@
...
+                    ?: // has no item on such position
+                    return false
+                return itemMatcher.matches(viewHolder.itemView)
+            }
+        }
     }
 
+
+
+    /* OLD TESTS USING FIREBASE -> COMMENTED TO BE REUSED WHEN FINAL DB IS PROPERLY SETUP */
```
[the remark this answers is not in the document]
Mouniraki: I did it this way because the Firebase database isn't completely setup for the moment, and the main goal of this task was to link the converstions from the ChatMenuActivity to the actual chat view. When everything will be properly setup, we will refactor this class to use Firebase.

app/src/main/java/com/github/brugapp/brug/model/Item.kt
Mouniraki: This is to be clarified for future sprints, but the way I designed my Conversation class was that I wanted to have the User with which we are talking to, the list of ChatMessage holding every message sent/received in the conversation, and the Item related to the discussion (i.e. if Anna has found my AirPods, I thought that it would be nice to remind me that she is related to my AirPods and not another item that I may also have lost).

omarelmalki: I see, thanks for your answer.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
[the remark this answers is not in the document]
omarelmalki: I see, thanks for your answer.

pmcintyr (approved):
Build.gradle)

coreLibraryDesugaring:
 'com.android.tools:desugar_jdk_libs:1.1.5'

On GitHub I found that the latest version is 1.2.0-alpha01. Do we specifically need v1.1.5 to enable LocalDateTime functions on older API level?


ChatMenuActivity)

Thank you for the comments in ChatMessagesListAdapter and in ChatActivityTest, they really helped me understand the code more easily. Would it be possible to add some documentation for the ChatMenuActivity? 

ChatMenuViewModel)

This is a small detail, so not necessary just an idea. For the ChatMenuViewModel, we could have more clarity and simplicity if we store all the sample messages in a ConversationSamples class. Then we could just call a shorter method from the new class to shorten the mock conversation list in the view model.

PR #116, opened 2022-03-24
Add notification reception and display

omarelmalki:
Add the ability to receive FCM Messages.
Add notification display both on demand and on reception of FCM message.
Now, the app can receive notifications from Firebase and display them even if it is running the background and notifications can also be displayed manually by clicking on the REPORT button.


app/src/main/res/layout/notification.xml
```diff
@@ -0,0 +1,41 @@
...
+            android:src="@drawable/unlost_logo" />
+
+        <TextView
+            android:id="@+id/notification_title"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginLeft="10dp"
+            android:layout_marginTop="15dp"
+            android:layout_toRightOf="@id/unlost_logo"
+            android:text="kskskskks"
```
ha2san, line 26: what this text mean ? If this mean nothing can we put an empty string instead ?

app/src/main/res/layout/notification.xml
```diff
@@ -0,0 +1,41 @@
...
+            android:textStyle="bold" />
+
+        <TextView
+            android:id="@+id/notification_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_below="@id/notification_title"
+            android:layout_marginLeft="10dp"
+            android:layout_toRightOf="@id/unlost_logo"
+            android:text="sidjkfkr"
```
ha2san, line 37: Same as what I said in the other comment, is it not better to put something meaningful or an empty string

app/build.gradle
```diff
@@ -41,12 +41,18 @@ android {
...
+    }
 }
 jacoco {
     toolVersion = "0.8.7"
 }
 
 dependencies {
+
+
+//    implementation 'androidx.work:work-runtime-ktx:2.5.0'
```
DeepGreen1, line 55: Why did you keep this comment here? Is it needed?

ha2san (approved):
Really nice work, it's a great start for implementing the notification in our app. I put some comment, I let you see if you think that the changes as necessary. I had just one question: what FCM means ?

app/src/androidTest/java/com/github/brugapp/brug/QrCodeScannerActivityTest.kt
```diff
@@ -33,7 +38,13 @@ class QrCodeScannerActivityTest {
             .check(matches((withHint("Report item…"))))
     }
 
-
+    @Test
+    fun reportButtonDisplaysNotificationWithoutCrashing(){
+        onView(withId(R.id.buttonReportItem))
+            .perform(ViewActions.click())
+        NotificationManagerCompat.from(ApplicationProvider.getApplicationContext()).cancelAll()
```
DeepGreen1, line 45: Should you not check here that the notification actually appeared?
Same remark for the ReportItemTest

app/src/main/res/layout/notification.xml
omarelmalki: Yes, you are right, it is was a placeholder to see how does the layout look. I'll delete it.

app/src/main/res/layout/notification.xml
omarelmalki: See previous comment.

omarelmalki:
> Really nice work, it's a great start for implementing the notification in our app. I put some comment, I let you see if you think that the changes as necessary. I had just one question: what FCM means ?

FCM means Firebase Cloud Messaging, we use it to send notifications and messages to users.

app/src/main/java/com/github/brugapp/brug/messaging/MyFCMMessagingService.kt
```diff
@@ -0,0 +1,89 @@
...
+                PendingIntent.FLAG_IMMUTABLE)
+
+            val channelId = context.getString(R.string.default_notification_channel_id)
+            val channelName = context.getString(R.string.default_notification_channel_name)
+
+            val defaultSoundUri = RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION)
+            val iconBitmap = BitmapFactory.decodeResource(context.resources, R.drawable.unlost_logo)
+            val notificationBuilder = NotificationCompat.Builder(context, channelId)
+                .setSmallIcon(R.drawable.ic_launcher_foreground)
+                .setLargeIcon(iconBitmap)
```
DeepGreen1, line 60: Nice work for this class! One idea for future improvement could be to store the Ringtone and vibration values somewhere so that the user can change it.

app/src/main/res/layout/notification.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <RelativeLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:layout_alignParentStart="true"
+        android:layout_alignParentTop="true"
+        android:padding="10dp">
+
+        <ImageView
+            android:id="@+id/unlost_logo"
```
DeepGreen1, line 14: Don't forget to check the XML editor remarks, here you could add the contentDescription for the ImageView. (It's useful for the blind people by eg.)

There is also some remarks about margin for right-to-left layouts and hard-coded strings.

app/build.gradle
```diff
@@ -41,12 +41,18 @@ android {
     buildFeatures {
         viewBinding true
     }
+    packagingOptions {
+        exclude "META-INF/DEPENDENCIES"
+    }
```
Mouniraki, line 46: Could you explain what does this statement ?

app/src/androidTest/java/com/github/brugapp/brug/ReportItemTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    @get:Rule
+    var qrCodeScannerActivityRule = ActivityScenarioRule(ReportItemActivity::class.java)
+
+    @Test
+    fun reportButtonDisplaysNotificationWithoutCrashing(){
+        onView(withId(R.id.report_item_button))
+            .perform(ViewActions.click())
+        NotificationManagerCompat.from(ApplicationProvider.getApplicationContext()).cancelAll()
+        Thread.sleep(1000)
```
Mouniraki, line 23: Same as last comment

app/src/main/java/com/github/brugapp/brug/ui/QrCodeScannerActivity.kt
```diff
@@ -31,4 +37,9 @@ class QrCodeScannerActivity : AppCompatActivity() {
         super.onPause()
     }
 
+    private fun displayReportNotification() {
+        MyFCMMessagingService.sendNotification(this, "Item found",
+            "One of your Items was found !")
+    }
+
```
Mouniraki, line 44: For next sprint you should put this method outside the Activities, since you are duplicating it in many different activities

app/src/main/java/com/github/brugapp/brug/ui/ReportItemActivity.kt
```diff
@@ -0,0 +1,24 @@
...
+
+        findViewById<Button>(R.id.report_item_button).setOnClickListener {
+            displayReportNotification()
+        }
+    }
+
+    private fun displayReportNotification() {
+        MyFCMMessagingService.sendNotification(this, "Item found",
+            "One of your Items was found !")
+    }
```
Mouniraki, line 23: See last comment

DeepGreen1 (commented):
Really nice work! I let you read the comments that I wrote. If you could at least update the XML issues, that would be perfect (quick to fix and it would remove lots of remarks for later reviews)

Concerning the test, I don't know if it's hard to do; it's up to you to change it now or in a future commit. Again, well done!

Mouniraki (approved):
Very good job, consider my comments below for the following sprint, PR approved

app/build.gradle
omarelmalki: I used this dependency during the development, it is not needed anymore. Thanks for the catch.

app/src/main/java/com/github/brugapp/brug/ui/QrCodeScannerActivity.kt
omarelmalki: The message would be different depending on the activity, I am only calling a one-line function in two activities, and their arguments will change.

app/build.gradle
omarelmalki: This statement excludes duplicate dependencies from the project that made the build fail.

app/src/androidTest/java/com/github/brugapp/brug/QrCodeScannerActivityTest.kt
omarelmalki: I cannot do that yet, with Espresso at least and it seems to be impossible to do in CI yet. 
This test, as its name indicate, checks that clicking the button to display the notification does not make the app crash.
Testing the content of the notification seems to be impossible in CI.
I am planning on using Mockito to mock Notifications in the tests to check if the output title and body is correct but that will still not solve this issue.

app/src/main/res/layout/notification.xml
omarelmalki: I tried following the right-to-left suggestions but it changed the result I was looking for.
Thanks for the suggestions

PR #117, opened 2022-03-24
Added backend functionality to show added items

hamzahass:


ha2san (approved):
Nice work, Just one question, in the Item class I saw that you have added an id parameter, have you an idea how we will compute it in the future. Because I see that we don't really use it right now.

app/src/androidTest/java/com/github/brugapp/brug/ItemTest.kt
```diff
@@ -28,7 +28,7 @@ class ItemTest {
         val itemName = "Wallet"
         val itemId = 1
```
Mouniraki, line 29: Instead of duplicating this statement in each of your test cases, you could set it as a constant or a `private val` as a class attribute

hamzahass:
> Nice work, Just one question, in the Item class I saw that you have added an id parameter, have you an idea how we will compute it in the future. Because I see that we don't really use it right now.

Thank you, so for the id I have left it in the MockDatabase as 0 for all items for now, as we will generate new ids for items for the database. For the how part we still need to figure out how to integrate the Items in the database and generate the ids for them, this will be in a separate sprint.

Mouniraki (approved):
Great job, take my comment into account for next time you will have write tests which involve a variable you use in many test cases.

app/src/androidTest/java/com/github/brugapp/brug/ItemTest.kt
hamzahass: Thanks for pointing this out, I will make sure to follow this!

PR #118, opened 2022-03-24
Feature/declared as lost

ha2san:
-Add feature to declare an item as lost
-Icon is now set automatically by the type of the item
-Remove iconId from Item constructor


omarelmalki (approved):
Great job on the declared as lost feature. Make sure you use the repository instead of the mock database in future sprints once it is implemented.

DeepGreen1 (approved):
Nice work! Your code is fine for me, I also checked the app on a phone and have no problem to report.
For the next steps, it could be nice to add a label "is lost" on the item list view so that the user don't have to go inside the menu of each item so see which one is declared as lost. Also, caching the list of item on firebase could be a good idea.

Again, well done!

PR #134, opened 2022-03-29
Integrate FirebaseInterface in RegisterUserViewModel

pmcintyr:
add extra Firebase helper functions
test account creation & authentication in FirebaseHelperTest

### DIFF
--- a/app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
+++ b/app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
@@ -1,5 +1,10 @@
 package com.github.brugapp.brug.data
 
+import android.content.ContentValues
+import android.util.Log
+import android.view.View
+import android.widget.ProgressBar
+import android.widget.Toast
 import com.google.android.gms.tasks.Task
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.auth.FirebaseUser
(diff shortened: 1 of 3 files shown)


DeepGreen1 (approved):
Nice work! Only two things to say:
- Try to add more documentation so that it's easier to understand what method does what when using the API (eg: createNewRegisterUser only creates a new user to register but that we have to call another method to insert it in firebase).
- It would be nice to address all the comments generated by Cirrus. They doesn't come from your commit but as they stack up with time it becomes cumbersome. (This review seems to be very long but it's actually mostly cirrus comments)

Again: well done!

pmcintyr:
> Nice work! Only two things to say:
> 
> * Try to add more documentation so that it's easier to understand what method does what when using the API (eg: createNewRegisterUser only creates a new user to register but that we have to call another method to insert it in firebase).
> * It would be nice to address all the comments generated by Cirrus. They doesn't come from your commit but as they stack up with time it becomes cumbersome. (This review seems to be very long but it's actually mostly cirrus comments)
> 
> Again: well done!

Hey! Thank you for your code review, I have added more precise documentation in the RegisterUserViewModel class in [the following commit.](https://github.com/BrugApp/brug/pull/134/commits/56bef77a393de8a097cec86d343cb4444c1db7b1).
I agree with you that we should address the Cirrus warnings, let's look into this if we have time after our current sprint tasks.

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
```diff
@@ -63,4 +68,65 @@ class FirebaseHelper {
...
+    /*
+    Returns a Task<DocumentReference> that tries to create a new FireBase User
+    Toasts text corresponding to the task's success/failure
+    @param  emailtxt  the email of the user that we are building
+    @param  passwordtxt the password of the user that we are building
+    @param  firstnametxt  the first name of the user that we are building
+    @param  lastnametxt the last name of the user that we are building
+    @return the Task<DocumentReference> that tries to create a new FireBase User
+    */
+    fun createAuthAccount(context: android.content.Context, progressBar: ProgressBar, emailtxt: String, passwordtxt: String, firstnametxt: String, lastnametxt: String){
```
Mouniraki, line 101: You can import the `android.content.Context` module to trim your code a bit

Mouniraki (approved):
Great ! Consider my comment before merging the code, but I also had a question:
Why do you have a different ViewModel class ? Because in it you only do calls to FirebaseHelper functions, what was the reason to have also a ViewModel class, instead of performing all necessary calls from the the SignIn Activities directly to the FirebaseHelper functions ?

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
pmcintyr: Thank you for your code review! I have a different ViewModel class in case we want to update the UI of the registration activity independently from the FirebaseHelper functions, this can be a good task idea in a future sprint!
I have optimized the `android.content.Context` import in [the latest commit.](https://github.com/BrugApp/brug/pull/134/commits/960fb0d91c270187cf6f2ff68ac9a33b18a8dec6)

PR #135, opened 2022-03-29
Add ER Model & Wiki entry to Repo

Mouniraki:
Addresses issue #122.
- Add new ER Model to illustrate our Firebase layout in the Wiki
- Add new entry to the Wiki to explain the Firebase structure (link : https://github.com/BrugApp/brug/wiki/Guide-to-Firebase---Firestore-Architecture)
- Implemented the structure in Firebase, Ready To Use as of now


pmcintyr (approved):
Hey good job with what you did here, here are some questions regarding the ER model:

1) Where would the qr-code fit within the model, would it be an attribute of item? does every item have a qr-code?

2) Does a message always take a ISA-child value or can it be instantiated itself without an ISA-child?

3) What is the icon_uid property for the User, does it identify the icon? Does userID not suffice for the icon mapping?

Mouniraki:
> Hey good job with what you did here, here are some questions regarding the ER model:
> 
>     1. Where would the qr-code fit within the model, would it be an attribute of item? does every item have a qr-code?
> 
>     2. Does a message always take a ISA-child value or can it be instantiated itself without an ISA-child?
> 
>     3. What is the icon_uid property for the User, does it identify the icon? Does userID not suffice for the icon mapping?

Thank you for your review ! Here are my answers to your questions:
1. Since every object associated to a user has a unique Item ID, the idea would be not to store the QR-Code as an image or byte-stream in the database, but to generate it in the app side when the user wants to order a sticker/plastic key holder. This would reduce the amount of data to be stored in the database.
2. In the Wiki I stated that because Firebase doesn't enforce that every "document" (a.k.a. entry) of a given collection (a.k.a. table) must have the same fixed set of attributes (i.e., the attributes of an entity do not belong to a collection but rather to a document of this very collection), we can represent every kind of message as an entry in the "message" sub-collection of the "Conversations" collection, where a VoiceMessage for example would simply have an additional `audio_url` attribute that a TextMessage wouldn't have. In our way of interacting with the database, however, we must make sure that every kind of message has at least the `sender`, `timestamp` and `body` attributes when we add a new message to the database.
3. Thank you for pointing out the little mistake, this attribute should reference the actual URL of the icon instead of another ID. I will make sure to correct that in the next commit.

DeepGreen1 (approved):
Nice work!
We could think about adding an email attribute for the Users and also maybe an attribute "has unread notifications" (think about having lost an item but someone scan the QR code before you enable notification for this item: it would be nice to be able to access them).
As it's up for discussion, I'll aprove your PR.

Mouniraki:
> Nice work! We could think about adding an email attribute for the Users and also maybe an attribute "has unread notifications" (think about having lost an item but someone scan the QR code before you enable notification for this item: it would be nice to be able to access them). As it's up for discussion, I'll aprove your PR.

Thanks ! After discovering the fact that there is already a Authentication database in Firebase holding the email and password of the registered users (or only the email address if the user signed in with his Google account), I decided that having the e-mail address of the user stored as an attribute in the Users relation doesn't make anymore sense, because it is already stored and linked to the User ID somewhere.
The "has unread notifications" should be discussed with everyone in my opinion, as we must clarify how we handle notifications in the situation you described in your comment (i.e., do we necessarily follow the `is_lost` state of the Item or do we do something else).

PR #143, opened 2022-04-02
Add User Item Repository and Fix Jacoco and CI

omarelmalki:
Fixed Jacoco and CI to recognize tests in test/ folder.
Adapted Item class to have String IDs for more compatibilty.
Added User Item Repository (feel free to uncomment and adapt/remove them in your code @pmcintyr when you finalize the repo).


app/src/main/java/com/github/brugapp/brug/data/UserItemRepository.kt
```diff
@@ -0,0 +1,98 @@
...
+//                } else {
+//                    Log.d(TAG, "No such document")
+//                }
+//            }
+//            .addOnFailureListener { exception ->
+//                Log.d(TAG, "get failed with ", exception)
+//            }
+//        return item
+//    }
+}
```
Mouniraki, line 98: Why did you comment the whole class content ? Will it be deleted in a future commit ?

Mouniraki (changes requested):
Good job, mainly refactorings as I see but consider my comment related on the UserItemRepository class. Also, since I see that you worked with the Firebase API, did you have issues to correctly retrieve data and create the User object, since everything is asynchronous ?

app/src/main/java/com/github/brugapp/brug/data/UserItemRepository.kt
omarelmalki: The content of the class conflicts with what @pmcintyr is currently working on as I needed it during my work. By commenting it, I leave the choice of keeping/adapting the code to him like we have discussed together after the last scrum meeting.

hamzahass (approved):
Good job on fixing the issue related to Unit tests not being recognized by Jacoco. However I have two questions : 
- How is making the item id a string better for compatibility ?
- But also (as Mounir mentioned) why did you comment all the UserItemRepository class ?

Mouniraki (approved):
Approving the PR from your explanations :)

omarelmalki:
> Good job on fixing the issue related to Unit tests not being recognized by Jacoco. However I have two questions :
> 
> * How is making the item id a string better for compatibility ?
> * But also (as Mounir mentioned) why did you comment all the UserItemRepository class ?

* In our database (Firebase) item ids are strings.
* I commented it to leave @pmcintyr the control over what is available in public in the repositories to avoid duplicate functions. We have already discussed that together.

PR #153, opened 2022-04-04
return local types in firebasehelper

pmcintyr:
refactor helper tasks to return local User & Item types as seen in our model & architecture

### DIFF
--- a/app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
+++ b/app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
@@ -52,15 +52,7 @@ class FirebaseHelper {
             var inputStream : Uri? = null
             var profilePicture: Drawable? = null
 
-            /* THIS IS HOW TO CONVERT A URI TO A DRAWABLE FOR UI CLASSES
-            try {
-                inputStream = uri?.let { getContentResolver().openInputStream(it) }
-                profilePicture = Drawable.createFromStream(inputStream, uri.toString())
-            } catch (e: Exception) {
-                print("uri to drawable conversion failed")
-            }*/
-
-            if (email == null || id == null || uri == null) { //    ||profilePicture == null
+            if (email == null || id == null || uri == null) {
                 return null
             } else {
                 return User(firstname, lastname, email, id, profilePicture)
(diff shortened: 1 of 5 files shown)


Mouniraki (approved):
Good job, but it seems that your coverage is very low, can you explain why is it like that ?

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
```diff
@@ -1,73 +1,130 @@
...
+            val id = mAuth.uid
+            val uri = mAuth.currentUser!!.photoUrl
+            var inputStream : Uri? = null
+            var profilePicture: Drawable? = null
 
-    //returns the collection of chats from the Firestore database
-    fun getChatCollection(): CollectionReference {
-        return db.collection("Chat")
-    }
+            /* THIS IS HOW TO CONVERT A URI TO A DRAWABLE FOR UI CLASSES
```
DeepGreen1, line 55: Are those comments (and the other few in this file) needed? Some of them seems to be "hints" that you used during devlopment.

DeepGreen1 (approved):
Great work! I saw that you began to write actual documentation, that really nice! There is some commented parts that you could maybe remove but everything else seems fine to me.
I agree with Mounir's question, maybe consider adding a task to implement tests in the next sprints.
Again, nice job!

pmcintyr:
> Good job, but it seems that your coverage is very low, can you explain why is it like that ?

I completely agree with you that the test coverage should be higher. All methods in this class have one or more tests, however _getCurrentUser_() and _getItemFromCurrentUser_() both have very high complexity. **I aim to further test these methods in the next sprint.**

One reason we have low coverage is because of impossible paths through the if-statements: getCurrentUser() cannot get a valid user with invalid parameters in practice because we do not allow Users to be created that way, however Firebase attribute calls can fail independently of their parent document.

app/src/main/java/com/github/brugapp/brug/data/FirebaseHelper.kt
pmcintyr: These comments are indeed unnecessary, I have removed them [in the latest commit ](https://github.com/BrugApp/brug/pull/153/commits/f36ba4b953dae95c82c52cfcf9befbdee0f3d006)

PR #159, opened 2022-04-07
Feature/voice messages

hamzahass:


app/build.gradle
```diff
@@ -57,6 +57,13 @@ dependencies {
     def lifecycle_version = "2.5.0-alpha05"
     def arch_version = "2.1.0"
 
+    //for AppCompat use:
+    //appcompat v26+ is highly recommended to support older APIs
+    //implementation 'com.devlomi.record-view:record-view:2.0.1'
```
pmcintyr, line 62: Is this commented record-view import going to be useful in future implementations or can we remove it?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -121,6 +124,84 @@ class ChatActivityTest {
...
+        val intent = Intent(context, ChatActivity::class.java).apply {
+            putExtra(CHAT_INTENT_KEY, conversation)
+        }
+
+        val message = "Test text"
+
+        ActivityScenario.launch<Activity>(intent).use {
+            onView(withId(R.id.editMessage)).perform(typeText(message))
+            onView(withId(R.id.recordButton)).check(matches(not(isDisplayed())))
+            onView(withId(R.id.editMessage)).perform(typeText(""))
```
pmcintyr, line 201: When we perform a typetext of an empty string does it click the editMessage and open the keyboard? Is this the reason why the record button turns off?

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -168,4 +200,119 @@ class ChatActivity : AppCompatActivity() {
             viewModel.uploadImage(this)
         }
     }
+
+    /* Function to test if device has a microphone,
```
pmcintyr, line 204: Are these commented methods functional already or are they going to be worked on more in the future? Is it ok to keep them here?

pmcintyr (approved):
Thanks for the voice message development! I have given you some comments/questions about the code. Also, you can assign the time labels and link the development issues!

app/build.gradle
hamzahass: This is for the having a nice UI when pressing the record button, basically just like whatsapp, however I can't manage to make it work for now, so I left it in order to continue working on that

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
hamzahass: I removed this on my last commit, but basically when we type something on the message text field, the record button disappears and the send message button appears.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
hamzahass: This was added for devices that don't have a microphone, but I don't know if it is pertinent, this is why I commented this

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -276,4 +281,64 @@ class ChatViewModel : ViewModel() {
             })
     }
     */
+
+    /* Will be uncommented when we figure out how to allow permissions in tests
```
ha2san, line 285: What is the exact issue ? Does the test initially pass on your machine ? We need to discuss this more deeply during our next meeting.

ha2san (approved):
Nice job, We need to find a way to also store the audio in the database to be able to retrieve them. As  I think that keeping them locally is not enough

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
hamzahass: The tests pass on my machine, as I manually allowed permissions for audio and external storage. However on cirrius these tests fail, and I figured it was because of the permissions that were not granted on CI emulator, I still need to find a way to make that work, and we should definitely discuss that.

PR #160, opened 2022-04-07
ChatActivity testing

DeepGreen1:
Adds tests to improve code coverage


app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -143,16 +128,21 @@ class ChatViewModel : ViewModel() {
...
-        val newMessage = LocationMessage("Me", DateService.fromLocalDateTime(LocalDateTime.now()), LocationService.fromAndroidLocation(location).toString(), LocationService.fromAndroidLocation(location))//ChatMessage(locationString, 0, LocalDateTime.now(), "Location")
+        //TODO: Properly initialize new messages in messageresponse wrapper
+        //TODO: Add location on Firebase
+        //val locationString = "longitude: ${location.longitude}; latitude: ${location.latitude}"
+        val newMessage = LocationMessage(
+            "Me",
+            DateService.fromLocalDateTime(LocalDateTime.now()),
+            LocationService.fromAndroidLocation(location).toString(),
+            LocationService.fromAndroidLocation(location)
+        )//ChatMessage(locationString, 0, LocalDateTime.now(), "Location")
```
Mouniraki, line 139: Why do you put the textual content of the message as the string representation of the LocationService object ? It should be the content of the text-box if the user has entered some text before sending his location to make more sense, no ?

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -216,83 +204,25 @@ class ChatViewModel : ViewModel() {
...
-
         // resize image and display the new created one
         val resizedUri = resize(activity as ChatActivity, imageUri)
-        val newMessage = PicMessage("Me", DateService.fromLocalDateTime(LocalDateTime.now()), "An image", resizedUri.toString())
+        val newMessage = PicMessage(
+            "Me",
+            DateService.fromLocalDateTime(LocalDateTime.now()),
+            "An image",
+            resizedUri.toString()
+        )
```
Mouniraki, line 214: Same here, shouldn't the text contain something else than "An image" ? It would be better to have the content of the message below the image if some text has been entered

Mouniraki (changes requested):
Good job, but from the results of Code Climate I see that you only got 51% of diff-coverage, which is too low to be considered as acceptable, given that the PR aims at increasing the coverage by adding more tests.
Try to aim at least 75-80% of diff-coverage to make it ready to be merged to the main branch.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -224,87 +257,215 @@ class ChatActivityTest {
...
+        val context = ApplicationProvider.getApplicationContext<Context>()
+
+        Intents.init()
+
+        val intent = Intent(context, ChatActivity::class.java).apply {
+            putExtra(CHAT_INTENT_KEY, conversation)
+        }
 
+        ActivityScenario.launch<Activity>(intent).use {
+            val expectedIntent: Matcher<Intent> = allOf(hasAction(MediaStore.ACTION_IMAGE_CAPTURE))
```
pmcintyr, line 312: Why did we opt for MediaStore as our storage module? As you can see here, mediastore is deprecated: https://google.github.io/modernstorage/mediastore/ .
When we have time I would recommend upgrading to the following modern module: [modernstorage-storage](https://google.github.io/modernstorage/storage/)

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -224,87 +257,215 @@ class ChatActivityTest {
...
+        Intents.init()
+
+        val intent = Intent(context, ChatActivity::class.java).apply {
+            putExtra(CHAT_INTENT_KEY, conversation)
+        }
+
+        ActivityScenario.launch<Activity>(intent).use {
+            val messagesList = onView(withId(R.id.messagesList))
+
+            val expectedIntent: Matcher<Intent> = allOf(hasAction(MediaStore.ACTION_IMAGE_CAPTURE))
```
pmcintyr, line 353: same comment as earlier regarding the use of mediastore

pmcintyr (approved):

pmcintyr:
> Good job, but from the results of Code Climate I see that you only got 51% of diff-coverage, which is too low to be considered as acceptable, given that the PR aims at increasing the coverage by adding more tests. Try to aim at at least 75-80% of diff-coverage to make it ready to be merged to the main branch.

Hey Mounir! I think that codeclimate is giving a bad representation of JE's diff-coverage, when you look at the total-coverage(the only important coverage metric on which we are graded) you will see that an increase of +1.9% change, thus reaching 84% total coverage. This means that this commit is more "acceptable" than even our current main branch.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: Not actually. I'm currently working on a new feature that displays the location as an image (a map showing where you are). Thus there will be no message associated to the location.

Your remark only makes sense for pictures and it's another feature that we may add later to the project.

Also, this PR was ment to be for testing, so I'll not change it here.

DeepGreen1: Not actually. I'm currently working on a new feature that displays the location as an image (a map showing where you are). Thus there will be no message associated to the location.

Your remark only makes sense for pictures and it's another feature that we may add later to the project.

Also, this PR was ment to be for testing, so I'll not change it here.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: As said above, we could do that but it's another feature. For the time being, there is no text displayed with an image and I we still need to have a content to create a PicMessage, hence "an image" seemed apropriate.

DeepGreen1:
> > Good job, but from the results of Code Climate I see that you only got 51% of diff-coverage, which is too low to be considered as acceptable, given that the PR aims at increasing the coverage by adding more tests. Try to aim at at least 75-80% of diff-coverage to make it ready to be merged to the main branch.
> 
> Hey Mounir! I think that codeclimate is giving a bad representation of JE's diff-coverage, when you look at the total-coverage(the only important coverage metric on which we are graded) you will see an increase of +1.9% change, thus reaching 84% total coverage. This means that this commit is more "acceptable" than even our current main branch.

I agree with Paul. Also, in this PR, I mainly added tests so the code coverage should not be a problem.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -304,15 +223,39 @@ class ChatActivity : AppCompatActivity() {
...
+
+        val imageUri = data?.extras?.getString("imageUri")
+        if (imageUri != null) {
+            // this will be the case for the gallery image
+            // camera images returns null as extras
+            println("=== URI set ===")
+            viewModel.setImageUri(Uri.parse(imageUri))
+        }
+        viewModel.uploadImage(this)
+    }
```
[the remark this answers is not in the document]
DeepGreen1: Yes, that's a good idea! I'll think about it.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -1,61 +1,85 @@
...
+        GrantPermissionRule.grant(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
+
     @get:Rule
-    val permissionRuleExtStorage: GrantPermissionRule = GrantPermissionRule.grant(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    val permissionRuleCamera: GrantPermissionRule =
+        GrantPermissionRule.grant(android.Manifest.permission.CAMERA)
 
     private val convID = "0"
+
     //TODO: CHANGE THIS TO ACTUAL USER
```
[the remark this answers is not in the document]
DeepGreen1: Do you mean keeping this tests as it is (using a mock user) and adding a new test using firebase?
If yes, we can do that in a next sprint yeah!

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -104,15 +134,17 @@ class ChatActivityTest {
...
+                    atPosition(1, hasDescendant(withText(newMessageText)))
+                )
+            )
         }
     }
 
-    @Test
-    fun sendLocationCorrectlyAddsNewMessage(){
+    /*@Test
+    // Should be uncommented on the next merge (with the new location feature)
```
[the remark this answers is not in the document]
DeepGreen1: Yes! As Mounir is waiting on this PR to merge his job, I wanted to make it quick. This tests is failing due to a waiting time of Firestore but it's not so easy to solve. As I'm working on the new feature of the location this week, I'll add some tests and uncomment this one.

Mouniraki (approved):
After our discussion, since many of us depend on the merging of this PR to continue, I am approving it with the small coverage increase, but from now on, no PR with less than 75-80% of diff-coverage will be accepted to avoid a decrease in coverage in the whole project. Don't forget to do the proper testing for the location service in the related PR, I suggest you to add a new issue to avoid forgetting it if it is not already the case.

DeepGreen1:
Thanks. I'm dealing with the testing of location on my new PR for the new location feature (as said in the comment of the code), no worries. 
The total coverage is increasing, that's good! It's strange to have such diff-coverage tho, I agree.

PR #176, opened 2022-04-08
Feature/voice messages

hamzahass:
Re-allowed requesting permissions for audio and external storage on local emulator, which fixed bugs for cirrius.


ha2san (approved):
Nice, I see that you found a way to grant the permission. I think we should try to add some dependencies to be able to test everything. Other than that, everything seems to be ok

app/build.gradle
```diff
@@ -60,6 +60,13 @@ dependencies {
     def lifecycle_version = "2.5.0-alpha05"
     def arch_version = "2.1.0"
 
+    //for AppCompat use:
```
DeepGreen1, line 63: Are those comments needed?

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -41,11 +55,33 @@ class ChatActivity : AppCompatActivity() {
...
 
+        buttonSendTextMessage = findViewById(R.id.buttonSendMessage)
+
         initMessageList(model)
         initSendMessageButton(model)
         initSendLocationButton(model, locationManager, fusedLocationClient)
         initSendImageCameraButton(model)
         initSendImageButton(model)
+
+        //initSendImageCameraButton() // TODO: Implement this
```
DeepGreen1, line 66: I think that this method is implemented, why did you comment this line? If it stays like that, we will not able to send images using the camera.

DeepGreen1 (changes requested):
Nice work! I think that we do the same for other parts that uses permissions, It'll improve our code coverage.
I let you some comments and I'll let you see for the takeCameraImage function that is commented. Otherwise everything seems fine to me. Great job!

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
hamzahass: I don't understand this, as I did not comment this line, nor add the TODO, I will investigate on that ..

app/build.gradle
hamzahass: Yes, I will need this for the view of recording messages

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
hamzahass: After further investigation, this function call is made 3 lines above, and this TODO was surely added on my side when merging.

app/build.gradle
DeepGreen1: Alright! We keep it then ;)

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
DeepGreen1: No problem this was probably overlooked, we'll have to remove it anyway. Let's do it during next print as the ChatActivity will again be modified during next week.

DeepGreen1:
Alright, thanks for your answers. Everything seems fine to me!

DeepGreen1 (approved):

PR #189, opened 2022-04-22
Bind Firebase with App Objects

Mouniraki:
Addresses issues #164 and #175.

This PR involves the following changes:
- Split FirebaseHelper into many small repository classes
- Refactored activities to fetch data related to the signed in user from Firebase
- Re-routed the demo button to go to the ItemsMenuActivity right after signing in to the demo account of the app
- Removed dependencies to MockDatabase, to be removed in a future sprint
- Updated the wiki to reflect final database changes

Note : due to many issues related to the sign-in procedure with Google, I wasn't able to fully implement the retrieval of the current user. The app should be used only by going through the demo mode until the issue is fixed.

EDIT: Sign in with Google has been fixed, now fully functional :)


app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
```diff
@@ -77,47 +108,55 @@ class ChatMenuActivityTest {
...
-        chatList.perform(
-            RecyclerViewActions.actionOnItemAtPosition<ListViewHolder>(
-                0, GeneralSwipeAction(
-                    Swipe.SLOW, GeneralLocation.BOTTOM_RIGHT, GeneralLocation.BOTTOM_LEFT,
-                    Press.FINGER
-                )
-            )
-        )
-        snackBar.check(matches(withText(CHAT_CHECK_TEXT)))
+//        Thread.sleep(10000)
```
pmcintyr, line 111: Why is the sleep commented here and why was it used in the 1st place? If you no longer need it we can delete it, and if we need it I would love to understand why.

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
```diff
@@ -77,47 +108,55 @@ class ChatMenuActivityTest {
...
-        chatList.perform(
-            RecyclerViewActions.actionOnItemAtPosition<ListViewHolder>(
-                0, GeneralSwipeAction(
-                    Swipe.SLOW, GeneralLocation.BOTTOM_LEFT, GeneralLocation.BOTTOM_RIGHT,
-                    Press.FINGER
-                )
-            )
-        )
-        snackBar.check(matches(withText(CHAT_CHECK_TEXT)))
+//        Thread.sleep(10000)
```
pmcintyr, line 130: Same comment as mentioned for previous sleep.

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
```diff
@@ -77,47 +108,55 @@ class ChatMenuActivityTest {
...
     @Test
     fun clickOnItemGoesToChatActivity() {
-        Thread.sleep(10000)
-        val chatList = onView(withId(R.id.chat_listview))
-        chatList.perform(
-            RecyclerViewActions.actionOnItemAtPosition<ListViewHolder>(
-                0, click()
-            )
-        )
+//        Thread.sleep(10000)
```
pmcintyr, line 148: Same comment as mentioned for previous sleep.

pmcintyr (changes requested):
Hello, please take a look [at the following commit](https://github.com/BrugApp/brug/pull/180
). I already finished the FirebaseHelper repository classes refactoring on the 13/04/22, however I can see you also added the [repos 4 days later.](https://github.com/BrugApp/brug/commit/9ce2163f14b458cd63d9fc19d658ecc7fd0f217b).  Before this merge happens, I would appreciate it if I can first merge my weekly task: [Rename and abstract various local type repositories from Firebase helper #171](https://github.com/BrugApp/brug/issues/171).

I also noticed that you deleted my addDocumentMessage method as well as the addDocumentMessageTest method, I understand we changed how the firebase references work, but can't we keep a modified version of this method with your new references?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -48,26 +49,20 @@ class ChatActivityTest {
     @get:Rule
     val permissionRuleExtStorage: GrantPermissionRule = GrantPermissionRule.grant(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
 
-    private val convID = "0"
-    //TODO: CHANGE THIS TO ACTUAL USER
-    private val dummyUser = DummyUser("John", "Doe", File.createTempFile("tempIMG", ".jpg").path)
-    private val dummyItemName = "DummyItem"
+    private val dummyUser = MyUser("USER1", "Rayan", "Kikou", null)
 
     private val dummyDate = DateService.fromLocalDateTime(
```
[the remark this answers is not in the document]
Mouniraki: Thank you, the modification will be done in next commit.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -48,26 +49,20 @@ class ChatActivityTest {
     @get:Rule
     val permissionRuleExtStorage: GrantPermissionRule = GrantPermissionRule.grant(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
 
-    private val convID = "0"
-    //TODO: CHANGE THIS TO ACTUAL USER
-    private val dummyUser = DummyUser("John", "Doe", File.createTempFile("tempIMG", ".jpg").path)
-    private val dummyItemName = "DummyItem"
+    private val dummyUser = MyUser("USER1", "Rayan", "Kikou", null)
```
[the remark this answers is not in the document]
Mouniraki: I created this new type of User, since the one we have contains attributes and definitions that are useless in regards to how a User is represented in Firebase or how we access the different objects of a User from Firebase.
The new code using User instead of MyUser will have to be refactored to use this new MyUser class, since it will replace the User class after everything is merged properly (doing also a little name refactoring afterwards to make it more pertinent)

Mouniraki:
> Hello, please take a look [at the following commit](https://github.com/BrugApp/brug/pull/180). I already finished the FirebaseHelper repository classes refactoring on the 13/04/22, however I can see you also added the [repos 4 days later.](https://github.com/BrugApp/brug/commit/9ce2163f14b458cd63d9fc19d658ecc7fd0f217b). Before this merge happens, I would appreciate it if I can first merge my weekly task: [Rename and abstract various local type repositories from Firebase helper #171](https://github.com/BrugApp/brug/issues/171).
> 
> I also noticed that you deleted my addDocumentMessage method as well as the addDocumentMessageTest method, I understand we changed how the firebase references work, but can't we keep a modified version of this method with your new references?

Thank you for your review. Since my part relied on having the app fully linked to the Firebase database, I had to have repo classes holding the different functions to query the database. However, since your PR didn't pass the tests (and still doesn't), I couldn't wait on you to update the failing tests and instead did the split myself. But if you manage to fix your tests and pass the CI check, the best thing would be to merge your part before mine.

I don't recall removing the addDocumentMessage method and its related test case, but if I did it is surely because either I had a similar method already which was maybe a bit more complete (handling also upload of files to Firebase Storage for example), or because it didn't make sense to still have it in the repo.

app/src/main/java/com/github/brugapp/brug/ui/ItemInformationActivity.kt
```diff
@@ -1,39 +1,57 @@
...
         setTextAllView(textSet)
 
         val switch: SwitchCompat = findViewById(R.id.isLostSwitch)
         switch.isChecked = item.isLost()
         switch.setOnCheckedChangeListener { _, isChecked ->
-            viewModel.setLostValue(isChecked)
-        }
+            item.changeLostStatus(isChecked)
+            liveData(Dispatchers.IO){
+                emit(ItemsRepo.updateItemFields(item, Firebase.auth.currentUser!!.uid))
```
ha2san, line 40: Instead of calling firebase, we can change the cached value and send all the query when we quit the app for example

app/src/main/java/com/github/brugapp/brug/ui/ItemsMenuActivity.kt
```diff
@@ -48,12 +57,19 @@ class ItemsMenuActivity : AppCompatActivity() {
         return super.onOptionsItemSelected(item)
     }
 
-    private fun initItemsList(model: ItemsMenuViewModel){
+
+    // ONLY GETS THE LIST OF ITEMS RELATED TO THE USER, NOT THE FULL USER PROFILE !
```
ha2san, line 61: Same as before, isnt it better to keep the user information in a static variable, then we do not need to ask firebase everytime

ha2san (approved):
Really nice work !!!
I just have one concern, Isn't it better to have a static object "currentUser" and do all the query when we start the app. Doing so allow use to gain some performance by not asking firebase for data everytime but by getting the data from the static object. We can for example update the database only if the currentUser was modified, by checking after a certain amount of time.

Mouniraki:
> Really nice work !!! I just have one concern, Isn't it better to have a static object "currentUser" and do all the query when we start the app. Doing so allow use to gain some performance by not asking firebase for data everytime but by getting the data from the static object. We can for example update the database only if the currentUser was modified, by checking after a certain amount of time.

Thank you for your review ! Indeed, this would be a great idea, but I think that it should be a standalone task by itself for a future sprint, since it is more linked to the caching part which will be tackled after. I will add a new task entry so that we don't forget it :)

pmcintyr (approved):
Great job, thanks for your corrections, in my opinion we are good to go!

PR #196, opened 2022-04-24
Feature/chat activity location

DeepGreen1:


DeepGreen1:
Currently not able to test as the conversations are unavailable on firestore.

DeepGreen1:
UI is working as expected. I'll merge changes of last week before implementing the API calls for the map.

app/src/main/java/com/github/brugapp/brug/model/ChatMessagesListAdapter.kt
```diff
@@ -55,21 +66,31 @@ class ChatMessagesListAdapter(private val viewModel: ChatViewModel, private val
         val message: Message = messageList[position]
```
pmcintyr, line 66: I think that with some documentation this method could be a bit more clear so that we can improve readability

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -44,6 +53,7 @@ class ChatActivity : AppCompatActivity() {
     private lateinit var deleteAudio: ImageButton
     private lateinit var textMessage: EditText
 
+    private val simpleDateFormat = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.FRENCH)
```
pmcintyr, line 56: why do we use 2 different date formats here and in the other comment i made? is this necessary?

app/src/main/java/com/github/brugapp/brug/model/ChatMessagesListAdapter.kt
```diff
@@ -55,21 +66,31 @@ class ChatMessagesListAdapter(private val viewModel: ChatViewModel, private val
...
+    class ViewHolder(itemView: View, listener: onItemClickListener) :
+        RecyclerView.ViewHolder(itemView) {
+        init {
+            itemView.setOnClickListener {
+                listener.onItemClick(adapterPosition)
+            }
+        }
+
         private fun formatDateTime(date: LocalDateTime): String {
             val formatter = DateTimeFormatter.ofPattern("dd/MM/yy - HH:mm")
```
pmcintyr, line 87: why do we use 2 different date formats here and in the other comment i made? is this necessary?

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -92,34 +102,67 @@ class ChatViewModel : ViewModel() {
...
+                Manifest.permission.ACCESS_COARSE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED && ContextCompat.checkSelfPermission(
+                activity.applicationContext,
+                Manifest.permission.ACCESS_FINE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED
+        ) {
+            requestPermissions(
+                activity,
+                arrayOf(
+                    Manifest.permission.ACCESS_COARSE_LOCATION,
```
pmcintyr, line 123: can you explain to me what coarse_location and fine_location corresponds to for our project and why both are needed? thanks!!

pmcintyr (approved):
Overall the code seems correct, I have some suggestions for next sprint/questions that you can see below.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
DeepGreen1: This one was use with the mock database; it was not removed when firebase was linked to everything. This is a good idea for some refactoring thanks!

app/src/main/java/com/github/brugapp/brug/model/ChatMessagesListAdapter.kt
DeepGreen1: This one is the one that we use for the chat (see other answer above).

app/src/main/java/com/github/brugapp/brug/model/ChatMessagesListAdapter.kt
DeepGreen1: Yes!I think that it can be very useful to have documentation for every public classes of our repo. I'll see to add that in a next week.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -249,6 +267,41 @@ class ChatActivity : AppCompatActivity() {
         })
     }
 
+    @SuppressLint("NewApi")
+    fun createFakeImage(): Uri? {
```
hamzahass, line 271: What is this method for, some additional documentation would help !

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -92,34 +102,67 @@ class ChatViewModel : ViewModel() {
...
         // Clear the message field
         textBox.text = ""
+
+        adapter.notifyItemInserted(messages.size - 1)
     }
 
-    fun sendPicMessage(activity: ChatActivity, convID: String, picURI: Uri) {
+    fun sendPicMessage(activity: ChatActivity, convID: String) {
         val textBox = activity.findViewById<TextView>(R.id.editMessage)
+        //val resizedUri = resize(activity, imageUri) //still useful??
```
hamzahass, line 152: Is this commented line necessary ?

hamzahass (approved):
Nice work. Consider my two comments bellow. Furthermore, why is the image for the localisation blue, and there is not actually the localisation

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: The coarse_location gives you the position of the phone using wifi and the antenas. The fine_loaction uses the GPS system to provide the location (thus very precise). We need at least the fine location for our project as we want to help people to find each other (and not be in the same town by eg.)

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
DeepGreen1: This method is used to create and store an image (like the stub for the tests). As said above, documentation will be added on the whole project.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: No, Mounir renamed resize() in compressImage(). I kept it to see if every call to resize was changed as merging can cause issues of that sort sometimes. I'll remove it, thanks!

DeepGreen1:
> Nice work. Consider my two comments bellow. Furthermore, why is the image for the localisation blue, and there is not actually the localisation

Well, the message is considered as a location (when you click on it, you can see that a toast appears; this is not the case for the other messages).
I wanted to use the same API than Omar, as we discussed during the last meeting. The GPS coordinate is already stored, I'll just have to generate an image using the API and display it instead of the blue square. Now, why is it blue? It's a way to show the work that has been done.

PR #201, opened 2022-04-26
Add mapbox navigation

omarelmalki:
Added map menu with 3 main features:
- display items on a map
- location tracking
- navigation from current location to destination with car

Setup secret token in CI to allow Mapbox to work even on CI. To make it work on your side, I have sent you the private key to add to your project.


hamzahass (approved):
Good work! Works well on my end. However I would suggest adding a bit of documentation on the methods that are not overwritten. Furthermore, for the test, we discussed during the last meeting that there was an issue with a given key, did you find a solution for it, if not what are you planning to do ?

app/src/main/res/layout/activity_main.xml
```diff
@@ -8,14 +8,17 @@
 
     <TextView
         android:id="@+id/mainHelloWorld"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
+        android:layout_width="302dp"
+        android:layout_height="40dp"
```
Mouniraki, line 12: Be careful with the hardcoded bias values, it can lead to weird UI dispositions across different screen sizes.

build.gradle
```diff
@@ -12,12 +12,12 @@ plugins {
...
-        google()
-        maven { url 'https://jitpack.io' }
-    }
-}
+//allprojects {
+//    repositories {
+//        google()
+//        maven { url 'https://jitpack.io' }
+//    }
+//}
```
Mouniraki, line 20: Is this chunk of commented code temporary or will you use it in the future ? Consider putting a comment on top of it to explain why it is commented.

app/src/androidTest/java/com/github/brugapp/brug/NavigationTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        Intents.release()
+    }
+
+    @Test
+    fun clickingOnSeeItemsGoesToCorrectActivity() {
+        Espresso.onView(ViewMatchers.withId(R.id.seeItemsButton)).perform(ViewActions.click())
+
+        Intents.intended(
+            Matchers.allOf(
+                IntentMatchers.toPackage("com.github.brugapp.brug"),
```
Mouniraki, line 43: Try to store the package name in a global variable instead of copying it many times across test cases

Mouniraki (approved):
Great job ! Very nice code overall, consider my comments before merging.

app/src/main/res/layout/activity_main.xml
omarelmalki: Completely agree, I plan on going through UI in a future sprint and correct these kind of issues.

build.gradle
omarelmalki: No, this will not be used. I fixed a warning in the build as with the project configuration, this should be in settings.gradle.

app/src/androidTest/java/com/github/brugapp/brug/NavigationTest.kt
omarelmalki: Great idea, thanks for the suggestion !

build.gradle
omarelmalki: Deleted the comments in my last commit.

PR #204, opened 2022-04-28
Add bugfix related to the PicMessages

Mouniraki:
This PR aims at fixing a bug found by @DeepGreen1, where a message containing a picture would not send itself in the chat.
The culprit was on line 267, where the imageUri variable wasn't properly initialized, leading to an empty Uri when the user requested an image to be sent from the Gallery or the Camera.

EDIT: Another bug regarding the way the images were displayed has been fixed in the last commit, due to a missing call to the resize/compress function. 
EDIT2: Finally fixed the issue with the Camera, the culprit was lying in the onActivityResult method and was due to a weird behavior with the Camera, which didn't return the correct Uri and left the requested field as null instead after taking a picture.

### DIFF
--- a/app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
+++ b/app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
@@ -274,21 +274,5 @@ class ChatActivity : AppCompatActivity() {
                 viewModel.sendPicMessage(this, convID, viewModel.getImageUri())
             }
         }
-
-//        if (requestCode == TAKE_PICTURE_REQUEST_CODE && resultCode == RESULT_OK) {
-//
-//
-//
-//        } else if (requestCode == SELECT_PICTURE_REQUEST_CODE && resultCode == RESULT_OK) {
-//
-//        }
-//
-//        val imageUri = data!!.data ?: Uri.parse(data.extras?.getString(PIC_ATTACHMENT_INTENT_KEY))
-//        if (imageUri != null) {
-//            // this will be the case for the gallery image
-//            // camera images returns null as extras
-//            println("=== URI set ===")
-//            viewModel.sendPicMessage(this, convID, imageUri)
-//        }
     }
 }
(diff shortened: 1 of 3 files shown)


Mouniraki:
> Hi! Thank's for your update. I checked this fix but it seems that it only helps for the gallery images and not for the camera. I can patch this in my next PR so that the others don't need to merge twice (instead of doing it here). What do you think?

Could you provide me some more information on what doesn't work with the Camera part ? It would still not send the image ?
By the way, I also solved the issue where the images would be displayed either too small or too big.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -259,18 +258,21 @@ class ChatActivity : AppCompatActivity() {
...
-        val imageUri = data?.extras?.getString(PIC_ATTACHMENT_INTENT_KEY)
-        if (imageUri != null) {
-            // this will be the case for the gallery image
-            // camera images returns null as extras
-            println("=== URI set ===")
-            viewModel.sendPicMessage(this, convID, Uri.parse(imageUri))
+        if (resultCode == RESULT_OK){
+            if (requestCode == SELECT_PICTURE_REQUEST_CODE){
+                Toast.makeText(this, "Image selected", Toast.LENGTH_SHORT).show()
+                val imageUri = data!!.data ?: Uri.parse(data.extras?.getString(PIC_ATTACHMENT_INTENT_KEY))
```
hamzahass, line 264: Wouldn't be better to have this line in the viewModel ?

hamzahass (approved):
Thanks for fixing this bug, works well on my end. Please consider the comment that I left bellow

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
Mouniraki: For the moment we must do it like this due to the fact that we have to have the onActivityResult in the Activity instead of in the ViewModel, but since this method will have to be refactored to use the new way of handling results I keep it as a something we can try to do next.

DeepGreen1 (approved):
Now, it seems to be working, good work!
It's nice to have good dimensions for the images of the chat. Concerning the rest of the code, I'll change things a bit in the next PR as I also added a few fixes of the code (eg: for the tests).

PR #205, opened 2022-04-28
Feature/voice messages

hamzahass:


app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -273,7 +277,16 @@ class ChatViewModel : ViewModel() {
...
-    /* Will be uncommented when we figure out how to allow permissions in tests
+    fun sendAudio(){
+
+        // TODO: modify this implementation to adapt it for Firestore
+        val audioMessage = AudioMessage("Me", DateService.fromLocalDateTime(LocalDateTime.now()),
+            "Audio", audioPath)
+
+        messages.add(audioMessage)
+        adapter.notifyItemInserted(messages.size - 1)
+    }
```
Mouniraki, line 288: I see that you created a new function that handles the sending procedure of audio messages, but it doesn't save the message in Firebase. Do you plan to make sure that the messages are correctly sent in Firebase (I made a function sendAudioMessage in the ChatViewModel also for you to help you with it) in this PR or do you plan to do it for next time ?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -206,6 +207,24 @@ class ChatActivityTest {
         }
     }
 
+    @Test
+    fun initialChatSetupAfterAudio(){
```
Mouniraki, line 211: As I can see from the content of the test, you are trying to ensure that when an audio message is sent, it is properly displayed in the Chat. If so, you maybe should rename your test so that we understand what it does without having to dig into it.

Mouniraki (approved):
Very nice job ! It's great to see so much progress made until now ! Consider my comments before merging your code to the main branch. Also, was it a choice of yours to have the background of the audio messages darker than the rest of the messages ?

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: I think that it's better to keep it for the next sprint. As all the methods are ready, it can be pretty quick to do, but a bug is always ready to hit us right before a meeting!

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -62,8 +62,8 @@ class ChatActivity : AppCompatActivity() {
         initMessageList(model)
         initSendMessageButton(model)
         initSendLocationButton(model, locationManager, fusedLocationClient)
-        initSendImageCameraButton(model)
-        initSendImageButton(model)
+        initSendImageCameraButton()
+        initSendImageButton()
 
         //initSendImageCameraButton() // TODO: Implement this
```
omarelmalki, line 68: Now that it is implemented, is this comment still needed?

omarelmalki (approved):
Hi, great job ! It seems you have two different layouts for the audio messages, what are you planning to do with the second one? See my other comments below too.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
hamzahass: No actually the this test verifies that when an audio is set everything is still good in the layout of the chat, this is why I called it initial-Chat-Setup-After-Audio

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
hamzahass: This comment is related to the images of the chat, which is from @DeepGreen1 's code, he probably removed it in the recent branch that he has.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
hamzahass: I join @DeepGreen1 on that

hamzahass:
> Hi, great job ! It seems you have two different layouts for the audio messages, what are you planning to do with the second one? See my other comments below too.

For now I have two layouts to see if in the future it would be better to have the second one or keep the first one, I will remove one of them in the next sprint.

DeepGreen1:
Nice work! I agree with the 2 others.
I quickly checked your code on a real phone and it seems that sometimes, sending the audio let the application in an illegal state. (I'll let you see: the mediaRecorder.stop() seems to be the culprit)
Have you checked on a phone/tablet?

Otherwise, I tried on the emulator and everything works well. Good job!!

hamzahass:
> Nice work! I agree with the 2 others. I quickly checked your code on a real phone and it seems that sometimes, sending the audio let the application in an illegal state. (I'll let you see: the mediaRecorder.stop() seems to be the culprit) Have you checked on a phone/tablet?
> 
> Otherwise, I tried on the emulator and everything works well. Good job!!

Thanks! I have tried on the emulator and an android tablette and everything seems to be working well, we can check later on your device if there are any issues, and if any fix that.

PR #231, opened 2022-05-02
Fix Serializable issue with Drawable

Mouniraki:
Addresses issue #230.
The issue was caused by the fact that Drawable is seemingly not a Serializable object, hence leading to errors when passed as extra of an Intent.
Please pull the content of the associated branch, test by yourself that it works by clicking on the conversation with the custom user icon, and confirm that the fix actually works.

The UI is a bit weird with custom images and should be fixed, but for now it is not crucial to the correctness of the app. Feel free to add a fix if you feel the need to do it.

EDIT: The UI issue has also been resolved with a temporary fix (hardcoding the dimensions of the users profile pictures displayed in the ChatMenuActivity)

### DIFF
--- a/app/src/androidTest/java/com/github/brugapp/brug/UserRepositoryTest.kt
+++ b/app/src/androidTest/java/com/github/brugapp/brug/UserRepositoryTest.kt
@@ -102,10 +102,10 @@ class UserRepositoryTest {
         Firebase.auth.signOut()
 
         assertThat(updatedUser, IsNot(IsNull.nullValue()))
-        assertThat(updatedUser!!.getUserIcon(), IsNot(IsNull.nullValue()))
-        assertThat(
-            updatedUser.getUserIcon()!!.intrinsicWidth / updatedUser.getUserIcon()!!.intrinsicHeight,
-            IsEqual(drawable.intrinsicWidth / drawable.intrinsicHeight))
+        assertThat(updatedUser!!.getUserIconPath(), IsNot(IsNull.nullValue()))
+//        assertThat(
+//            updatedUser.getUserIconPath()!!.intrinsicWidth / updatedUser.getUserIcon()!!.intrinsicHeight,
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/com/github/brugapp/brug/UserRepositoryTest.kt
```diff
@@ -102,10 +102,10 @@ class UserRepositoryTest {
...
 
         assertThat(updatedUser, IsNot(IsNull.nullValue()))
-        assertThat(updatedUser!!.getUserIcon(), IsNot(IsNull.nullValue()))
-        assertThat(
-            updatedUser.getUserIcon()!!.intrinsicWidth / updatedUser.getUserIcon()!!.intrinsicHeight,
-            IsEqual(drawable.intrinsicWidth / drawable.intrinsicHeight))
+        assertThat(updatedUser!!.getUserIconPath(), IsNot(IsNull.nullValue()))
+//        assertThat(
+//            updatedUser.getUserIconPath()!!.intrinsicWidth / updatedUser.getUserIcon()!!.intrinsicHeight,
+//            IsEqual(drawable.intrinsicWidth / drawable.intrinsicHeight))
```
hamzahass, line 108: Are these commented lines still necessary ?

hamzahass (approved):
Nice work, thanks for fixing this bug. I pulled your branch and it works well on my end, please consider the comment I left bellow.

app/src/androidTest/java/com/github/brugapp/brug/UserRepositoryTest.kt
Mouniraki: I left them as I wanted to keep them around in order to fix the final assertion, but I will decide on whether I keep it or not for the PR of my main task since the commented assertion might not be very useful anymore.

ha2san (approved):
Everything seems to work correctly. The bug is indeed fixed on my emulator and phone.

PR #237, opened 2022-05-05
Test/firebase emulator

ha2san:


ha2san:
Setting up Firebase emulator in the project in order to facilitate the tests and to reduce the execution time by avoiding queries in the network

ha2san:
You can see how to use the emulators by reading the [wiki](https://github.com/BrugApp/brug/wiki/Firebase-emulator)

omarelmalki (commented):
Great work on implementing the firebase emulator for tests.

I tried running the tests locally with `sudo firebase emulators:exec './gradlew connectedCheck'` but it cannot access a secret key stored on my machine.

My solution was to first run `sudo firebase emulators:start` and then in a new terminal run `./gradlew connectedCheck`. Is this normal?

Otherwise, the emulator seems to work correctly.

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
```diff
@@ -36,33 +43,62 @@ import org.junit.Test
...
 @RunWith(AndroidJUnit4::class)
 class AddItemTest {
 
+
     @get:Rule
-    var testRule = ActivityScenarioRule(AddItemActivity::class.java)
+    var rule = HiltAndroidRule(this)
+
+   // @get:Rule
+   // var testRule = ActivityScenarioRule(AddItemActivity::class.java)
```
hamzahass, line 57: Are these comments still necessary

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
```diff
@@ -249,7 +256,7 @@ class ChatActivityTest {
 
         ActivityScenario.launch<Activity>(intent).use {
             onView(withId(R.id.recordButton)).perform(click())
-            Thread.sleep(2000)
+            Thread.sleep(1000)
```
hamzahass, line 259: Why did you reduce the Thread sleep time on this, and do you know why was it initially added ?

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
```diff
@@ -30,46 +35,149 @@ import org.junit.Rule
...
+                test_user_uid,
+                firestore,
+                firebaseAuth,
+                firebaseStorage
+            )
+            //NO MESSAGE SO THERE IS AN ERROR
+            Log.d("createConv", result.onError.toString())
+            Log.d("convs list", convs.toString())
+        }
+
```
hamzahass, line 160: Why are you creating a new conversation for the tests, aren't conversations already present ?

hamzahass (approved):
Great work! I pulled your branch, followed the wiki and the firebase emulator works well on my end. However consider the comments I left bellow.

Mouniraki:
Hey ! I tried running the tests with Firebase Emulator setup on my computer, but the test suite fails on two tests, with those errors :
`com.github.brugapp.brug.UserRepositoryTest > updateUserIconToInexistentUserReturnsError[Pixel_5_API_31(AVD) - 12] FAILED 
        com.google.firebase.auth.FirebaseAuthInvalidUserException: There is no user record corresponding to this identifier. The user may have been deleted.
        at com.google.android.gms.internal.firebase-auth-api.zzti.zza(com.google.firebase:firebase-auth@@21.0.3:27)

com.github.brugapp.brug.sign_in.SignInActivityButtonTest > itemFoundButtonGoesToQRCodeScannerActivity[Pixel_5_API_31(AVD) - 12] FAILED 
        androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: view.getId() is <2131231265/com.github.brugapp.brug:id/qr_found_btn>
`

Did you have the same issue in your end at some point ?

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
ha2san: No these are not necessary, I'll get rid of them

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
ha2san: the emulator does not keep the conversation between each run so I have to add them

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
ha2san: this line is used to keep the button pressed for a while. I found that reducing the time did not harm the tests and saved us time.

ha2san:
> Hey ! I tried running the tests with Firebase Emulator setup on my computer, but the test suite fails on two tests, with those errors :
> 
> com.github.brugapp.brug.UserRepositoryTest > updateUserIconToInexistentUserReturnsError[Pixel_5_API_31(AVD) - 12] FAILED com.google.firebase.auth.FirebaseAuthInvalidUserException: There is no user record corresponding to this identifier. The user may have been deleted. at com.google.android.gms.internal.firebase-auth-api.zzti.zza(com.google.firebase:firebase-auth@@21.0.3:27)
> 
> com.github.brugapp.brug.sign_in.SignInActivityButtonTest > itemFoundButtonGoesToQRCodeScannerActivity[Pixel_5_API_31(AVD) - 12] FAILED androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: view.getId() is <2131231265/com.github.brugapp.brug:id/qr_found_btn>
> 
> Did you have the same issues in your end at some point ?


Hi, concerning the second error, it can happen sometimes, Rerun the test and tell me if it works. For the first one it is quite strange as I create the user before executing the test. I must have forgotten something. I'll get back at you when I have found the source of the error

Mouniraki:
> > Hey ! I tried running the tests with Firebase Emulator setup on my computer, but the test suite fails on two tests, with those errors :
> > com.github.brugapp.brug.UserRepositoryTest > updateUserIconToInexistentUserReturnsError[Pixel_5_API_31(AVD) - 12] FAILED com.google.firebase.auth.FirebaseAuthInvalidUserException: There is no user record corresponding to this identifier. The user may have been deleted. at com.google.android.gms.internal.firebase-auth-api.zzti.zza(com.google.firebase:firebase-auth@@21.0.3:27)
> > com.github.brugapp.brug.sign_in.SignInActivityButtonTest > itemFoundButtonGoesToQRCodeScannerActivity[Pixel_5_API_31(AVD) - 12] FAILED androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: view.getId() is <2131231265/com.github.brugapp.brug:id/qr_found_btn>
> > Did you have the same issues in your end at some point ?
> 
> Hi, concerning the second error, it can happen sometimes, Rerun the test and tell me if it works. For the first one I did not come across this error. It is quite strange as I create the user before executing the test. I must have forgotten something. I'll get back at you when I have found the source of the error

For the first error I think I have found the issue, in the AddItemTest you don't ensure that the statement adding a new user in FirebaseAuth completes before executing all the next instructions in the test, leading to an error in case the add statement takes too long to return. You should add an await() statement to it. Let me know when you will have the final version of the code pushed to the repo so that I can re-test and review afterwards :)

ha2san:
> Great work on implementing the firebase emulator for tests.
> 
> I tried running the tests locally with `sudo firebase emulators:exec './gradlew connectedCheck'` but it cannot access a secret key stored on my machine.
> 
> My solution was to first run `sudo firebase emulators:start` and then in a new terminal run `./gradlew connectedCheck`. Is this normal?
> 
> Otherwise, the emulator seems to work correctly.

this is quite strange as @hamzahass and @Mouniraki did not have this issue. I'll try to figure out where the problem can come frome

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
```diff
@@ -36,33 +43,59 @@ import org.junit.Test
...
 
+
     @get:Rule
-    var testRule = ActivityScenarioRule(AddItemActivity::class.java)
+    var rule = HiltAndroidRule(this)
+
+    private val firebaseAuth: FirebaseAuth = FirebaseFakeHelper().providesAuth()
+    private val firestore: FirebaseFirestore = FirebaseFakeHelper().providesFirestore()
+    companion object{
+        var first_time = true
```
Mouniraki, line 59: What does this variable mean here ? Is it some kind of flag to avoid having to instantiate too many times some modules while performing the tests ?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
Mouniraki: Did you test it without the Thread.sleep() there ? Just to know why it is still necessary :)

Mouniraki (approved):
Great job ! I see that all the files are also now re-formatted which is really nice. Consider my comments and try to re-run the whole procedure to check if the coverage increases or not, so that we can see together what causes the low coverage at times.

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
ha2san: This flag is here in order to not create two times an account with Firebase. As we are using Junit4, we do not have the @beforeAll flags in order to create the accounts before running the tests. I found that this is the simplest way to avoid this issue.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityTest.kt
ha2san: I'll try

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
Mouniraki: I see, thank you for your reply ! I think that a good thing to do then would be to put a little comment on top of the variable assignment to make it more clear to all of us.

PR #242, opened 2022-05-08
Fixed audio messages crash when sent

hamzahass:
This bug was related to the fact that the audio messages were stored on the external storage of the device (SD Card), which was okay in the case of the emulator, but for actual devices, not all of them have external storage.

### DIFF
--- a/app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
+++ b/app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
@@ -21,7 +21,6 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.FileProvider
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
-import com.devlomi.record_view.RecordButton
 import com.github.brugapp.brug.PIC_ATTACHMENT_INTENT_KEY
 import com.github.brugapp.brug.R
 import com.github.brugapp.brug.SELECT_PICTURE_REQUEST_CODE
@@ -49,7 +48,7 @@ class ChatActivity : AppCompatActivity() {
     private lateinit var convID: String
 
     private lateinit var buttonSendTextMessage: ImageButton
-    private lateinit var recordButton: RecordButton
+    private lateinit var recordButton: ImageButton
     private lateinit var messageLayout: LinearLayout
     private lateinit var audioRecMessage: TextView
     private lateinit var buttonSendAudio: ImageButton
@@ -82,7 +81,6 @@ class ChatActivity : AppCompatActivity() {
         buttonSendAudio = findViewById(R.id.buttonSendAudio)
 
         recordButton = findViewById(R.id.recordButton)
-        viewModel.setListenForRecord(recordButton, false)
         initRecordButton(viewModel)
 
         deleteAudio = findViewById(R.id.deleteAudio)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -266,9 +249,8 @@ class ChatActivity : AppCompatActivity() {
             audioRecMessage.visibility = View.GONE
             messageLayout.visibility = View.VISIBLE
             recordButton.visibility = View.VISIBLE
-            model.setListenForRecord(recordButton, false)
 
-            model.sendAudio()
+            model.sendAudio(this, convID, firestore, firebaseAuth, firebaseStorage)
```
pmcintyr, line 253: Hi, I saw your refactoring to merge setListenForRecord and sendAudio into 1 function. Are we sure we never want to use setListenForRecord without sendAudio? If this were the case, separating the methods could be an advantage. If not, no problem, just explain to me why this is the case.

app/src/test/java/com/github/brugapp/brug/MessageTest.kt
```diff
@@ -246,8 +247,9 @@ class MessageTest {
         )
 
         val audioUrl = "/path/to/dummy/audio.mp3"
+        val audioPath = "/path/to/dummy/audio.mp3"
```
pmcintyr, line 250: I was wondering if we need to keep both variables here(audioUrl & audioPath), considering they contain the same dummy string. Can't we just keep 1?

app/src/test/java/com/github/brugapp/brug/MessageTest.kt
```diff
@@ -263,9 +265,10 @@ class MessageTest {
         ))
         val body = "TESTMESSAGE"
         val audioUrl = "/path/to/dummy/audio.mp3"
```
pmcintyr, line 267: Same thing here for audioUrl & audioPath

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -55,7 +54,7 @@ class ChatActivity : AppCompatActivity() {
     private lateinit var convID: String
 
     private lateinit var buttonSendTextMessage: ImageButton
-    private lateinit var recordButton: RecordButton
+    private lateinit var recordButton: ImageButton
```
pmcintyr, line 57: Could you explain to me why we changed the Button type here? Thanks!!

pmcintyr (approved):

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
hamzahass: For the setListenForRecord function I just deleted for two reasons, the first being that I stopped using the RecordView layout for the audios (and setListenForRecord was I function of that library), but also we don't need to set for listen, as when we click on the button to record, the button disappears (visibility set to gone) and the buttons to stop and delete the audio show up.

app/src/test/java/com/github/brugapp/brug/MessageTest.kt
hamzahass: Actually we can, but it's just to have a better understanding of the code. In that case the url and the path are the same, but in some cases it isn't, so I figured it was better to keep them separate.

app/src/test/java/com/github/brugapp/brug/MessageTest.kt
hamzahass: The reason is the same as the other

ha2san (approved):
Really nice catch, this shows that just because everything works on the emulator does not mean that there are no problems. Well done

PR #246, opened 2022-05-10
QR-Code Scan-And-Notify Feature

Mouniraki:
Addresses issue #167.

**This PR involves the following changes :**
- Modified getters for Conversation and Message objects from Firebase repos to now have them re-trigger each time a new event occurs in Firebase
- Replaced the list of messages held by a Conversation by only the last message sent in the conversation, since now messages are updated independently of the conversation when a new message arrives
- Add new Conversation upon scanning a QR Code in QrCodeScannerActivity
- Send a LocationMessage with the location of the device when scanning the QR Code
- Send a generic TextMessage saying that an item has been found
- Implemented message cache to save messages, was required with new implementation
- ChatActivity and ChatMenuActivity can now receive a list of message/conversation from Intents, this was needed for the tests (not for the users)
- Users now hold a list of device tokens in order to be able to receive targeted notification messages
- Refactored code for sign-in, making sure that every operation performed in order before saving the device token to the user account

**Minor bugfixes:**
- Conversations holding no messages can now be displayed. This was due to a too harsh condition which was present in the Repo classes on the number of valid messages retrieved from Firebase.

**What remains to be implemented :**
- Notification handling (i.e. send a notification to a user when a modification to which he/she is related occurred in Firebase)


app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -71,28 +71,57 @@ class ConvRepositoryTest {
     @Test
     fun addNewConvCorrectlyReturns() = runBlocking {
         assertThat(ConvRepository.addNewConversation(USER_ID1, USER_ID2, DUMMY_ITEM_NAME,firestore).onSuccess, IsEqual(true))
-        val conversation = Conversation("${USER_ID1}${USER_ID2}", USER2, DUMMY_ITEM_NAME, mutableListOf())
-        val convList = ConvRepository.getUserConvFromUID(USER_ID1, firestore,firebaseAuth,firebaseStorage)
-        assertThat(convList.isNullOrEmpty(), IsEqual(false))
-        val conv = convList!!.last()
+        val conversation = Conversation("${USER_ID1}${USER_ID2}", USER2, DUMMY_ITEM_NAME, null)
+//        val convList = ConvRepository.getUserConvFromUID(USER_ID1, firestore,firebaseAuth,firebaseStorage)
```
hamzahass, line 75: Is this comment still needed

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -71,28 +71,57 @@ class ConvRepositoryTest {
...
+
+        val context = TestLifecycleOwner()
+        ConvRepository.getRealtimeConvsFromUID(USER_ID1, context, firestore, firebaseAuth, firebaseStorage)
+        delay(1000)
+        val convList = BrugDataCache.getConversationList()
+        assertThat(convList.value.isNullOrEmpty(), IsEqual(false))
+
+//        assertThat(convList.value.isNullOrEmpty(), IsEqual(false))
```
hamzahass, line 83: same for this

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -71,28 +71,57 @@ class ConvRepositoryTest {
...
+    fun getConvsFromNonexistentUserReturnsNull() {
+        BrugDataCache.resetConversationsList()
+        val context = TestLifecycleOwner()
+        val wrongConvID = "WRONGCONVID"
+        ConvRepository.getRealtimeConvsFromUID(wrongConvID, context, firestore, firebaseAuth, firebaseStorage)
+        runBlocking {
+            delay(1000)
+        }
+        assertThat(BrugDataCache.getConversationList().value, IsEqual(mutableListOf()))
+//        assertThat(ConvRepository.getUserConvFromUID("WRONGCONVID",firestore,firebaseAuth,firebaseStorage), IsNull.nullValue())
```
hamzahass, line 100: same here

app/src/androidTest/java/com/github/brugapp/brug/MessageRepositoryTest.kt
```diff
@@ -95,23 +96,45 @@ class MessageRepositoryTest {
...
+        val messagesList = BrugDataCache.getMessageList("${USER_ID1}${USER_ID2}")
+
+        assertThat(messagesList.value.isNullOrEmpty(), IsEqual(false))
+        assertThat(messagesList.value!!.contains(TEXTMSG), IsEqual(true))
+
+//        val conv = ConvRepository.getUserConvFromUID(USER_ID1,
+//            firestore,firebaseAuth,firebaseStorage)!!.filter {
+//            it.convId == "${USER_ID1}${USER_ID2}"
+//        }
+
```
hamzahass, line 137: For these comments too

hamzahass (approved):
Great work ! This was probably one of our last big features. Please first consider the comments I have left bellow, regarding some commented lines in your code, I guess that you refactored some code and maybe forgot to delete them. Secondly I would like to know what are you considering to do for the notifications, in the current situation, when a user scans an Item is a conversation directly launched with the owner, and just the notification isn't sent ?

DeepGreen1 (approved):
This is again a big piece of work, nice job! I checked that the new features where working on a real device (eg: a new message appears in the conversation when recieved) and everything works well.
I saw that you begun to write documentation, that's good but there is some comments that are left in the code (I let you check this). It would be great to remove what is not necessary.
You also made a lot of adaptations on the "old code" where there is less documentation: it's not your job for this sprint, but it would be great to add documentation also for that.
Again, well done!

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
```diff
@@ -226,15 +154,15 @@ class ChatMenuActivityTest {
 
         entryToSwipe.swipeLeft(50)
 
-        val snackBarTextView = device.findObject(UiSelector().resourceId(SNACKBAR_ID))
-        assertThat(snackBarTextView.text, IsEqual(CHAT_CHECK_TEXT))
+//        val snackBarTextView = device.findObject(UiSelector().resourceId(SNACKBAR_ID))
```
pmcintyr, line 157: Why are these lines commented? Are we still going to need the snackbarTextview and the corresponding assertion?

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: Thank you for pointing it out, it should be removed and will be with all the other mentioned comments before merging.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -39,20 +40,19 @@ private val ACCOUNT1 = BrugSignInAccount("Rayan", "Kikou", "", "")
...
+    private val firebaseStorage: FirebaseStorage = FirebaseFakeHelper().providesStorage()
 
     //NEEDED SINCE @Before FUNCTIONS NEED TO BE VOID
     private fun addTestUsers() = runBlocking{
-        UserRepository.addUserFromAccount(USER_ID1, ACCOUNT1,firestore)
-        UserRepository.addUserFromAccount(USER_ID2, ACCOUNT2,firestore)
-        UserRepository.addUserFromAccount(USERWRONGCONV_ID, ACCOUNTWRONGCONV,firestore)
+        UserRepository.addUserFromAccount(USER_ID1, ACCOUNT1, true, firestore)
+        UserRepository.addUserFromAccount(USER_ID2, ACCOUNT2, true, firestore)
+        UserRepository.addUserFromAccount(USERWRONGCONV_ID, ACCOUNTWRONGCONV, true, firestore)
```
pmcintyr, line 55: What was the disadvantage of the old method addAuthUser that led you to use addUserFromAccount?

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -39,20 +40,19 @@ private val ACCOUNT1 = BrugSignInAccount("Rayan", "Kikou", "", "")
...
 
 //        UserRepo.addAuthUser(USER2)
 //        UserRepo.addAuthUser(USERWITHWRONGCONV)
```
pmcintyr, line 58: You can delete these 2 comments in my opinion also

Mouniraki:
> Great work ! This was probably one of our last big features. Please first consider the comments I have left bellow, regarding some commented lines in your code, I guess that you refactored some code and maybe forgot to delete them. There are many other occurrences for commented lines, I suggest you quickly go over the code. Secondly I would like to know what are you considering to do for the notifications, in the current situation, when a user scans an Item is a conversation directly launched with the owner, and just the notification isn't sent ?

Thank you for your review ! Indeed, I have missed some commented lines and they should be removed, which will be done in the last commit before merging. For the notifications part, I wasn't able to fully make it work due to issues with FirebaseMessaging that I will discuss with the staff, so for now there are no notifications sent when a user receives a message or a new conversation is created. But, I have implemented all the necessary bits and pieces to make it work, it is only the calls to FirebaseMessaging which don't work for some reason.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -71,28 +71,57 @@ class ConvRepositoryTest {
...
+//        val convList = ConvRepository.getUserConvFromUID(USER_ID1, firestore,firebaseAuth,firebaseStorage)
+
+        val context = TestLifecycleOwner()
+        ConvRepository.getRealtimeConvsFromUID(USER_ID1, context, firestore, firebaseAuth, firebaseStorage)
+        delay(1000)
+        val convList = BrugDataCache.getConversationList()
+        assertThat(convList.value.isNullOrEmpty(), IsEqual(false))
+
+//        assertThat(convList.value.isNullOrEmpty(), IsEqual(false))
+        val conv = convList.value!!.last()
```
pmcintyr, line 84: Should we consider the case where convList.value is null? Maybe  we could assign something else conv in this case

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -71,28 +71,57 @@ class ConvRepositoryTest {
...
-        assertThat(ConvRepository.getUserConvFromUID(USERWRONGCONV_ID,firestore,firebaseAuth,firebaseStorage), IsEqual(listOf()))
+    fun getBadlyFormattedConvsReturnsEmptyList() {
+        BrugDataCache.resetConversationsList()
+        val context = TestLifecycleOwner()
+        ConvRepository.getRealtimeConvsFromUID(USERWRONGCONV_ID, context, firestore, firebaseAuth, firebaseStorage)
+        runBlocking {
+            delay(1000)
+        }
+        assertThat(BrugDataCache.getConversationList().value, IsEqual(mutableListOf()))
+//        assertThat(ConvRepository.getUserConvFromUID(USERWRONGCONV_ID,firestore,firebaseAuth,firebaseStorage), IsEqual(listOf()))
```
pmcintyr, line 112: I found 3 commented assertions in this class, if we can remove them it might help readability

app/src/androidTest/java/com/github/brugapp/brug/FirebaseAuthRepositoryTest.kt
```diff
@@ -25,7 +25,7 @@ class FirebaseAuthRepositoryTest {
         val dummyAccount = BrugSignInAccount("DUMMYFNAME", "DUMMYLNAME", "", TEST_USER_EMAIL)
         createAccout()
         //Why have we to create a account before ?
```
pmcintyr, line 27: I don't really understand this comment, i know it isn't from your last commit but do you understand it and could you explain it to me?

app/src/androidTest/java/com/github/brugapp/brug/MessageRepositoryTest.kt
```diff
@@ -95,23 +96,45 @@ class MessageRepositoryTest {
...
             firebaseStorage
-            )
-        assertThat(response.onSuccess, IsEqual(true))
+        )
 
-        val conv = ConvRepository.getUserConvFromUID(USER_ID1,
-            firestore,firebaseAuth,firebaseStorage)!!.filter {
-            it.convId == "${USER_ID1}${USER_ID2}"
+        runBlocking {
+            delay(1000)
```
pmcintyr, line 125: Is the delay needed here? Why or why not?

pmcintyr (approved):

app/src/androidTest/java/com/github/brugapp/brug/FirebaseAuthRepositoryTest.kt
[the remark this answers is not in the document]
Mouniraki: Yes you are right about this, but this wasn't part of my task and I think it was put in the Firebase Emulator PR. I will keep it as it may be useful for the next sprint in case of refactoring code related to this.

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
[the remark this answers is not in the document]
Mouniraki: Yes, they will be very useful for next sprint as I am planning to re-check every test and try to do more assertions.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: This way of doing it is better in my opinion, as the sign-in operation returns a SignInAccount object in our local implementation no matter how we sign-in (i.e by Google or Email or otherwise). Also, I wrote it using the Kotlin Coroutines, which is a much easier way to code asynchronous code and to understand its content when sharing it afterwards.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: You are right, those comments will be gone in my last commit.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: No since in this case it doesn't make sense, the test should check that when I am adding a Conversation to the database, this conversation should actually be added to the database and not something else, hence I must have that the conversation list is not empty or null.

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: I will check for the other two but the highlighted one will be removed, thanks :)

app/src/androidTest/java/com/github/brugapp/brug/FirebaseAuthRepositoryTest.kt
Mouniraki: I didn't add it, I think it is related to the Firebase Emulator PR and it seems to be important for the next sprint, so I will keep it there.

app/src/androidTest/java/com/github/brugapp/brug/MessageRepositoryTest.kt
Mouniraki: Yes, the delay is mandatory here, since the new way of retrieving messages isn't blocking due to Firebase's implementation of Coroutine functions, and I had to ensure that the conversation list is fully retrieved before doing all the next operations. Not ideal, but I had to do it that way.

app/src/androidTest/java/com/github/brugapp/brug/ChatMenuActivityTest.kt
Mouniraki: Same reply as for https://github.com/BrugApp/brug/pull/246#discussion_r871547649, click on the link to check what I already said.

app/src/androidTest/java/com/github/brugapp/brug/QrCodeScannerActivityTest.kt
```diff
@@ -6,46 +6,123 @@ import androidx.lifecycle.Lifecycle
...
 @RunWith(AndroidJUnit4::class)
 class QrCodeScannerActivityTest {
     @get:Rule
-    var qrCodeScannerActivityRule = ActivityScenarioRule(QrCodeScannerActivity::class.java)
+    var rule = HiltAndroidRule(this)
 
+    @get:Rule
+    var qrCodeScannerActivityRule = ActivityScenarioRule(QrCodeScannerActivity::class.java)
 
     //https://stackoverflow.com/questions/33929937/android-marshmallow-test-permissions-with-espresso
```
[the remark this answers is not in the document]
Mouniraki: This will be removed as it is no longer needed, thanks for pointing it out :)

PR #253, opened 2022-05-11
Feature/location api

DeepGreen1:


app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -362,17 +399,47 @@ class ChatViewModel : ViewModel() {
...
             "Audio", Uri.fromFile(File(audioPath)).toString(), audioPath
         )
         sendMessage(audioMessage, convID, activity, firestore, firebaseAuth, firebaseStorage)
     }
 
+    // HELPERS METHODS =======================================================
+    @SuppressLint("NewApi")
+    fun createFakeImage(activity: ChatActivity): Uri? {
+        val encodedImage =
+            "iVBORw0KGgoAAAANSUhEUgAAAKQAAACZCAYAAAChUZEyAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAAEnQAABJ0Ad5mH3gAAAG0SURBVHhe7dIxAcAgEMDALx4rqKKqDxZEZLhbYiDP+/17IGLdQoIhSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSFIMSYohSTEkKYYkxZCkGJIUQ5JiSEJmDnORA7zZz2YFAAAAAElFTkSuQmCC"
```
Mouniraki, line 420: I know that it is possible to retrieve an image from the assets of the app, isn't it a better idea to do so instead of generating an image with a very long string which seems random ?

Mouniraki (approved):
Good job ! Overall, nothing special to say, for next sprint we really should focus on how to rescale the images properly to avoid deformations, and also documentation, which is lacking a bit. Consider my comment before merging.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: Yes, you're right! I wanted to do so at first but had some issue to load the assets within the test. My workaround was to create this blue image from a string. I'll check again if it's possible to access the assets.

DeepGreen1:
> Good job ! Overall, nothing special to say, for next sprint we really should focus on how to rescale the images properly to avoid deformations, and also documentation, which is lacking a bit. Consider my comment before merging.

Thanks! Yes, for the next print, the idea is to focus on the UI & documentation. I'll see that the ratio of the images is kept when uploaded on the database. (If time I'll try to begin this tonight)

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -70,13 +75,27 @@ class ChatViewModel : ViewModel() {
...
         }
 
         this.adapter = ChatMessagesListAdapter(this, messages)
     }
 
+    private fun getUrlForLocation(activity: ChatActivity, location: Location): URL {
+        val lat = location.latitude.toString()
+        val lon = location.longitude.toString()
+        val baseUrl =
+            "https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/geojson(%7B%22type%22%3A%22Point%22%2C%22coordinates%22%3A%5B$lon%2C$lat%5D%7D)/"
```
omarelmalki, line 92: It would be great to save this as a saved string.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -70,13 +75,27 @@ class ChatViewModel : ViewModel() {
...
+        val posUrl = "$lon,$lat"
+        val endUrl =
+            ",15/500x500?logo=false&attribution=false&access_token=" + activity.getString(R.string.mapbox_access_token)
```
omarelmalki, line 95: same here

omarelmalki (approved):
Great work on the display of static location on the chat. It works well on my end. Check my comments below for improvements.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: You mean in the ressources? Actually I'm modifying it with the parameters lon and lat. I could save the first part in the resosources if it makes the code clearer tho, good idea!

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: Only a part of it could be saved in the ressources, but it's a good idea yes! I'll do this with the PR concearning the ratio of the images. Thanks

PR #285, opened 2022-05-15
Add last location to Firebase Items

omarelmalki:
Add last location to Firebase items when getting items from or updating items in Firestore.

Add item location to firebase when clicking on chat map.


app/src/main/java/com/github/brugapp/brug/data/ItemsRepository.kt
```diff
@@ -42,7 +43,9 @@ object ItemsRepository {
                     "item_name" to item.itemName,
                     "item_type" to item.itemTypeID,
                     "item_description" to item.itemDesc,
-                    "is_lost" to item.isLost()
+                    "is_lost" to item.isLost(),
+                    "last_latitude" to item.getLastLocation()?.getLatitude(),
+                    "last_longitude" to item.getLastLocation()?.getLongitude()
```
Mouniraki, line 48: I see that you set the last location by splitting the longitude and latitude in two distinct fields. However, Firebase has the GeoPoint type handling location objects, and we have a LocationService class handling these objects in our app. Consider using it instead of doing what you have done.

Mouniraki (changes requested):
Good job, but consider my comments before getting an approval. It will improve the readability/correctness of the app.

app/src/main/java/com/github/brugapp/brug/data/ItemsRepository.kt
omarelmalki: Good idea, done in last commit.

app/src/main/java/com/github/brugapp/brug/data/ItemsRepository.kt
```diff
@@ -42,7 +45,8 @@ object ItemsRepository {
                     "item_name" to item.itemName,
                     "item_type" to item.itemTypeID,
                     "item_description" to item.itemDesc,
-                    "is_lost" to item.isLost()
+                    "is_lost" to item.isLost(),
+                    "last_location" to item.getLastLocation()?.toFirebaseGeoPoint()
```
Mouniraki, line 49: Are you sure that setting a GeoPoint object to null in Firebase doesn't produce any errors ? In case there is no last location for the item, isn't it a better idea to simply not include the field in the Firebase document, and check if the field is present in the getter function ?

Mouniraki (approved):
Great job ! Please see my comments before merging, they can be very useful for improving readability & correctness of your functions I think.

app/src/main/java/com/github/brugapp/brug/data/ItemsRepository.kt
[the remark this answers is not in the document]
omarelmalki: This is an outdated version of the code. This issue was fixed a few commits ago.

app/src/main/java/com/github/brugapp/brug/ui/ChatActivity.kt
```diff
@@ -138,10 +144,16 @@ class ChatActivity : AppCompatActivity() {
...
                     adapter.getItemViewType(position) == ChatMessagesListAdapter.MessageType.TYPE_LOCATION_LEFT.ordinal
                 ) {
-                    val myIntent = Intent(this@ChatActivity, MapBoxActivity::class.java)
                     val message = adapter.getItem(position) as LocationMessage
-                    myIntent.putExtra(EXTRA_DESTINATION_LONGITUDE, message.location.toAndroidLocation().longitude)
-                    myIntent.putExtra(EXTRA_DESTINATION_LATITUDE, message.location.toAndroidLocation().latitude)
+                    val lon = message.location.toAndroidLocation().longitude
+                    val lat = message.location.toAndroidLocation().latitude
+                    liveData(Dispatchers.IO) {
+                        conversation.lostItem.setLastLocation(lon, lat)
```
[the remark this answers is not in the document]
omarelmalki: This could be done indeed by creating a new constructor but for now I am keeping the version with longitude and latitude.

app/src/main/java/com/github/brugapp/brug/data/ItemsRepository.kt
omarelmalki: This doesn't produce errors, I tried it. I believe it is a better idea to standardise attributes of items by keeping it null when the last location is not available.

app/src/androidTest/java/com/github/brugapp/brug/ItemsRepositoryTest.kt
```diff
@@ -128,4 +129,29 @@ class ItemRepoTest {
...
+    @Test
+    fun getUserItemsFromUIDTest() = runBlocking {
+        ITEM.setItemID(ITEM_ID)
+        ITEM.setLastLocation(0.0, 0.0)
+        val response = ItemsRepository.addItemWithItemID(ITEM, ITEM_ID, DUMMY_UID,firestore)
+        val list = ItemsRepository.getUserItemsFromUID(DUMMY_UID,firestore)
+
+        assertThat(response.onSuccess, IsEqual(true))
+        assertThat(list.isNullOrEmpty(), IsEqual(false))
+        assertThat(list!!.contains(ITEM), IsEqual(true))
```
hamzahass, line 141: Consider putting only one assertion for better maintainability, as it would be better to find the problem if the test fails in the futur

app/src/androidTest/java/com/github/brugapp/brug/ItemsRepositoryTest.kt
```diff
@@ -128,4 +129,29 @@ class ItemRepoTest {
...
+        ITEM.setItemID(ITEM_ID)
+        ITEM.setLastLocation(0.0, 0.0)
+        ItemsRepository.addItemWithItemID(ITEM, ITEM_ID, DUMMY_UID,firestore)
+        val updatedItem = MyItem("AirPods 3", 1, ITEM.itemDesc, ITEM.isLost())
+        updatedItem.setItemID(ITEM_ID)
+        assertThat(ItemsRepository.updateItemFields(updatedItem, DUMMY_UID,firestore).onSuccess, IsEqual(true))
+        val items = ItemsRepository.getUserItemsFromUID(DUMMY_UID,firestore)
+
+        assertThat(items.isNullOrEmpty(), IsEqual(false))
+        assertThat(items!!.contains(updatedItem), IsEqual(true))
```
hamzahass, line 155: same here

hamzahass (approved):
Thanks for adding the last localisation of objects to Firebase, consider my comments bellow.

PR #291, opened 2022-05-16
Update .gitignore

DeepGreen1:
Update of the gitignore in order to keep our own firebase emulator config files. (Depends on OS and device)


pmcintyr (approved):
After a long and deep reflection, I agree with these many changes

omarelmalki (approved):
Good job, this ensures that our local configurations do not overlap.

PR #292, opened 2022-05-16
Replace Item Name by Item ID

Mouniraki:
This PR is linked to issue #285.
It involves the following changes:
- Refactor Conversation to hold a MyItem object instead of only the lost item name
- Refactor ConvRepository to register an item ID instead of only the lost item name


app/src/main/java/com/github/brugapp/brug/ui/QrCodeScannerActivity.kt
```diff
@@ -41,7 +41,7 @@ class QrCodeScannerActivity : AppCompatActivity() {
         viewModel.codeScanner(this)
 
         //TODO: REMOVE THIS HARDCODED TEXT
-        findViewById<EditText>(R.id.edit_message)
+        findViewById<EditText>(R.id.edit_message).setText("84suOx4k0feEMUkjgEAeYrpWvlf1:GC1bZYp7AfVoQOo5u9uY")
```
omarelmalki, line 44: What does this text mean?

omarelmalki (approved):
Great job on making the item related to the conversation available.

app/src/androidTest/java/com/github/brugapp/brug/sign_in/SignInActivityTestFakeHandler.kt
```diff
@@ -170,13 +170,14 @@ class SignInActivityTestFakeHandler {
...
+//    @Test
+//    fun clickOnGoogleSignInButtonDoesNotCrash() {
+//        val intent = Intent(ApplicationProvider.getApplicationContext(), SignInActivity::class.java)
+//
+//
+//        ActivityScenario.launch<SignInActivity>(intent).use {
+//            Thread.sleep(1000)
+//            onView(withId(R.id.sign_in_google_button)).perform(click())
+//        }
+//    }
```
hamzahass, line 182: Is this test still needed ?

hamzahass (approved):
Nice work, please consider the comment I left bellow. Furthermore, why gig we make that change ?

app/src/main/java/com/github/brugapp/brug/ui/QrCodeScannerActivity.kt
Mouniraki: It is a sample userID:itemID string that I used to check whether or not the addConversation procedure worked correctly. I will leave it here to be able to check/debug issues that could come up with the QrCodeScannerActivity, for instance when we will merge the NFC part with the QR scanning part.

app/src/androidTest/java/com/github/brugapp/brug/sign_in/SignInActivityTestFakeHandler.kt
Mouniraki: Yes, in fact I commented it because it failed many times and didn't belong to my part, but I discussed about it to the concerned teammates and they will have a look into it to properly fix it.

PR #296, opened 2022-05-16
Added return buttons to activities + fixed navbar buttons bug

hamzahass:
This PR contains : 

- New Return buttons for activities
- Fixed the bug on the navigation bar when pressing the back button


omarelmalki (approved):
Great job, the fix works well on my end.

ha2san (approved):
thanks for the fix, it works well on my end

PR #304, opened 2022-05-17
Unify User Interface

omarelmalki:
Unify user interface with a black and white theme.


ha2san (approved):
really nice work. the new color theme suits me. The change made in the SettingsActivity are really nice. I just have one issue with the two scanning button in the main menu, they are touching each other. I have another issue with the EditText in the QrCodeScanner, I feel like the line of text has become smaller.

DeepGreen1 (approved):
Nice work, it changes a lot from the pruple version!
However, there is some changes and remarks that have to be made:

- The home menu have buttons that collides with another; it would be nice to have the "Google button" with a same visual than the other buttons
- The AddItem activity was not updated with this new design (we discussed it and it'll be changed in the next PR)
- The messages from ChatActivity have the image that is swaped (I'll update this in a next PR)
- Do you plan to change the recording animation in the ChatActivity?
- Launching the Map activity result in a crash of the app

Otherwise, everything looks good! Also, I tried to set the night mode on a real device, the colors are changed! That's good, but there is still some work that can be done here (also in a next PR, as discussed).

I'll let you update the home menu buttons and fix the crash, then good job!

PR #308, opened 2022-05-17
Complete cache implementation

Mouniraki:
Addresses issue #324.
- Added new functions/attributes in cache to hold all the remaining types of objects in the app
- Changed the method of fetching items, now done in real-time like Conversation and Message objects (to comply with new feature saving last found location inside items)
- Fixed issues related to MapBox, where for instance the fetching of the image crashed if a loss of connectivity occurred
- Removed drag-to-reorder feature, doesn't make sense anymore due to constraints in number of Firebase queries since we added the new behavior with items
- Removed old Item and User classes, replaced by the new MyItem and MyUser classes (renamed accordingly to replace the old ones)
- Add a message informing the user when a loss of connectivity occurred

**EDIT: A bug in MapBox was found, where some items may not be displayed correctly sometimes, which makes the MapBox tests crash. A fix is in the works, but will not be part of this PR.**


app/src/main/java/com/github/brugapp/brug/view_model/ItemsMenuViewModel.kt
```diff
@@ -37,51 +42,44 @@ class ItemsMenuViewModel : ViewModel() {
...
-                    )
-                )
-            }.observe(activity) { response ->
+            if(!isTest){
+                viewModelScope.launch {
+                    val result = ItemsRepository.deleteItemFromUser(
+                        deletedItem.getItemID(),
+                        firebaseAuth.currentUser!!.uid,
+                        firestore
+                    ).onSuccess
```
ha2san, line 51: Why do we have to make a difference between testing and not testing mode

ha2san (approved):
really nice work, thank you for your big work.
Can you please respond to my comment.

app/src/main/java/com/github/brugapp/brug/view_model/ItemsMenuViewModel.kt
Mouniraki: It's because the tests for the ItemsMenuActivity now only take a list of dummy items which are not linked to Firebase, hence a statement like this one would fail if it is executed and as a consequence I had to make this statement executable only if we are not in a testing scenario in ItemsMenuActivityTest.

app/src/androidTest/java/com/github/brugapp/brug/ItemsMenuActivityTest.kt
```diff
@@ -186,46 +159,46 @@ class ItemsMenuActivityTest {
...
+//
+//        val snackBarTextView = device.findObject(UiSelector().resourceId(SNACKBAR_ID))
+//        assertThat(snackBarTextView.text, IsEqual(ITEMS_MOVE_TEXT))
+//    }
+//
+//    @Test
+//    fun dragDownOnItemReordersList(){
+//        val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+//
+//        val itemsList = UiScrollable(UiSelector().resourceId(LIST_VIEW_ID))
```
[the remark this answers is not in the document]
Mouniraki: I left the tests here because I wanted to discuss whether or not we should remove the drag-to-reorder feature from the list of items, since the order is never saved in Firebase.

omarelmalki (changes requested):
Great job with the offline mode finalisation. I tested the app on my end by disabling internet access at different moments and I noticed a few issues:

1. When there is no connection  and user is not signed in:
Fails when clicking on button:
- Sign In WIth Google in SignInActivity -> crash
- Report in QRCodeScanActivity-> E/FirebaseMessaging: Failed to get FIS auth token -> crash

2. Activating connection while it is running sometimes throws a :E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-3

3. When there is no connection and user is signed in:
-  QrCodeScan Report -> crash

4. When connection is activated after sign in:
- click on map button on botton nav bar sometimes crashes the app

Connection is available:
- sometimes clicking on Demo runs indefinitely without stopping, it would be great to have a timeout that warns the user that sign in was not possible.

Mouniraki:
> Great job with the offline mode finalisation. I tested the app on my end by disabling internet access at different moments and I noticed a few issues:
> 
>     1. When there is no connection  and user is not signed in:
>        Fails when clicking on button:
> 
> 
>     * Sign In WIth Google in SignInActivity -> crash
> 
>     * Report in QRCodeScanActivity-> E/FirebaseMessaging: Failed to get FIS auth token -> crash
> 
> 
>     2. Activating connection while it is running sometimes throws a :E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker
> 
>     3. When there is no connection and user is signed in:
> 
> 
>     * QrCodeScan Report -> crash
> 
> 
>     4. When connection is activated after sign in:
> 
> 
>     * click on map button on bottom nav bar sometimes crashes the app
> 
> 
> Connection is available:
> 
>     * sometimes clicking on Demo runs indefinitely without stopping, it would be great to have a timeout that warns the user that sign in was not possible.

Thank you for your review ! Indeed, I was able to reproduce the issues and have fixes for almost all of them, except the Map issue. Since you were in charge of the MapBox implementation in the app, it would be great if we could check the issue together and try to work our way towards a fix. The other fixes will be pushed very soon.

app/src/main/java/com/github/brugapp/brug/ui/MapBoxActivity.kt
```diff
@@ -98,27 +97,44 @@ class MapBoxActivity : AppCompatActivity() {
...
+        if(itemsTestList == null){
+            // First it fetches the items from Firebase, so that the list of items is fresh
+            ItemsRepository.getRealtimeUserItemsFromUID(firebaseAuth.uid!!, this, firestore)
+        } else {
+            BrugDataCache.setItemsInCache(itemsTestList)
+        }
+
+        val annotationPlugin = binding.mapView.annotations
+        val pointAnnotationManager = annotationPlugin.createPointAnnotationManager()
+        val viewAnnotationManager = binding.mapView.viewAnnotationManager
```
omarelmalki, line 114: These can only be set once in the beginning and re-used afterwards. I would recommend adding them as activity variables and initializing them in the onCreate method.

omarelmalki (approved):
Great work. I have been able to re-test all the issues I have and they are fixed now. As discussed in private when we were actively working on fixing a bug with the MapBoxAcivity tests as a result to the addition of the cache, the issue seems to be fixed on my end too with the latest changes and there is no need to comment these tests anymore.

On another note, please check my comment below for possible improvements.

app/src/main/java/com/github/brugapp/brug/ui/MapBoxActivity.kt
Mouniraki: Yes you are right, these should be initialized at the start of the activity since they are only handlers. This will be made in a following commit.

PR #310, opened 2022-05-18
Feature/image ratio

DeepGreen1:


app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -263,9 +262,31 @@ class ChatViewModel : ViewModel() {
         return image
     }
 
+    fun computeWidthHeight(width: Int, height: Int, targetWidth: Int, targetHeight: Int): Pair<Int, Int>{
+        val ratioBM = width.toFloat() / height.toFloat()
```
pmcintyr, line 266: Hello, why do we use floats here rather than Ints?

pmcintyr (approved):

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -263,9 +262,31 @@ class ChatViewModel : ViewModel() {
...
+        val idealWidth = targetWidth.toFloat()
+        val idealHeight = targetHeight.toFloat()
+
+        var resizedWidth = idealWidth
+        var resizedHeight = idealHeight
+
+        // the width and height will never exceed 500 by keeping ratio
+        if (ratioBM > 1){
+            // the width is larger than the height and ratio > 1
+            resizedHeight = idealHeight / ratioBM
```
ha2san, line 276: same here, maybe check that ratioBM != 0

ha2san (approved):
Nice, it works on my end. But please pay attention to my comment

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
```diff
@@ -204,7 +204,6 @@ class ChatViewModel : ViewModel() {
     ) {
         val textBox = activity.findViewById<TextView>(R.id.editMessage)
         val strMsg = textBox.text.toString().ifBlank { "📷 Image" }
-        //val resizedUri = resize(activity, imageUri) //still useful??
         val newMessage = PicMessage(
             "Me",
             DateService.fromLocalDateTime(LocalDateTime.now()),
```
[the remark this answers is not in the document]
DeepGreen1: Yes, that could be a good idea but as I only use it here, I thought that would not be usefull.

app/src/main/java/com/github/brugapp/brug/view_model/ChatViewModel.kt
DeepGreen1: Hello, it's because I think that it makes more sense as we are displaying images in pixel size. Also, the method to resize from Bitmap uses integers.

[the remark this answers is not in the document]
DeepGreen1: Yeah, nice catch! However, in the worst case we would have a null image (and thus an exception would be thrown) but I think that having an image of height or width of 0 is not likely to happen. I'll think about it thanks!

PR #311, opened 2022-05-19
Testing/tests je

DeepGreen1:


build.gradle
```diff
@@ -16,4 +16,3 @@ plugins {
 task clean(type: Delete) {
     delete rootProject.buildDir
 }
```
omarelmalki, line 18:
```suggestion
}

```
I would suggest re-adding this blank line back to solve your issues with Cirrus CI.

omarelmalki (commented):
Checkout my comment below to see if it solves your issue with Cirrus CI.

build.gradle
DeepGreen1: Thanks! I had to to it for both build.gradle files, what a sneaky bug!

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
```diff
@@ -0,0 +1,142 @@
...
+                .perform(ViewActions.click())
+
+            // deny permissions
+            pressOnPermission("Manifest.permission.ACCESS_FINE_LOCATION")
+
+            Espresso.onView(ViewMatchers.withId(R.id.buttonSendLocalisation))
+                .perform(ViewActions.click())
+
+            // deny permissions (again)
+            pressOnPermission("Manifest.permission.ACCESS_FINE_LOCATION")
```
pmcintyr, line 115: Hey, good job with the thorough testing, this will likely help us attain an even better total test coverage. I was wondering why we deny the permissions twice here?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
```diff
@@ -0,0 +1,142 @@
...
+                .perform(ViewActions.click())
+
+            // deny permissions
+            pressOnPermission("Manifest.permission.CAMERA")
+
+            Espresso.onView(ViewMatchers.withId(R.id.buttonSendImagePerCamera))
+                .perform(ViewActions.click())
+
+            // deny permissions (again)
+            pressOnPermission("Manifest.permission.CAMERA")
```
pmcintyr, line 139: same question as above here, why do we use 2 permission denials?

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.time.LocalDateTime
+import java.time.Month
+
+@HiltAndroidTest
+@RunWith(AndroidJUnit4::class)
+class ChatActivityPermissionDeniedTest {
```
pmcintyr, line 37: Hi, why did we choose to test permissions in a separate class than in chatActivitytest? Is it something to do with the available rules?

app/src/androidTest/java/com/github/brugapp/brug/SettingsActivityTest.kt
```diff
@@ -1,30 +1,49 @@
...
 class SettingsActivityTest {
     @get:Rule
-    var mainActivityRule = ActivityScenarioRule(SettingsActivity::class.java)
+    var rule = HiltAndroidRule(this)
 
     @get:Rule
-    var rule = HiltAndroidRule(this)
+    var settingsActivityRule = ActivityScenarioRule(SettingsActivity::class.java)
+
+    // is always failling on Cirrus (passes on local)
```
pmcintyr, line 34: Do we want to keep this test if it fails on Cirrus? Maybe we should consider deleting it.

pmcintyr (approved):

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
DeepGreen1: Thanks! I wanted to check the behavior of our application when the user denies the permission twice (the message is different and we should ask the permission again).

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
DeepGreen1: Same answer :)

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
DeepGreen1: This is due to the GrantPermissionRule. When using it, the permission is given to the application (it basically bypasses the permissions; no UI is displayed). When granted however, there is no possbility to revoke it unless you restart all the tests. I tried to use orchestra that have a flag to clear the memory of the application between each tests, but then Jacoco and Codeclimate have some issues as they keep track of the coverage in a file that is deleted with the memory. Hence, the only option that I found was to keep the tests with all permissions in one file and to tests the denials in another one (where we have to explicitly grant each permission using the UI). Hopes it clarifies!

app/src/androidTest/java/com/github/brugapp/brug/SettingsActivityTest.kt
DeepGreen1: Yes, I wanted to check the code coverage with the modifications of everybody. If we still want to improve it, I'll work on this test (it works on my machine so I just have to figure out why it's failling on Cirrus).

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        )
+    )
+
+    // adapted from https://gist.github.com/rocboronat/65b1187a9fca9eabfebb5121d818a3c4
+    private val PERMISSIONS_DIALOG_DELAY = 2000L
+    private val GRANT_BUTTON_INDEX = 1 // 0 to accept, 1 to deny
+
+    private fun pressOnPermission(permissionNeeded: String?) {
+        try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !hasNeededPermission(
```
hamzahass, line 68: I am not sure to understand what his if clause does, will this test pass only for certain Build.VERSION.SDK_INT ?

hamzahass (approved):
Good work on adding tests for the ChatActivity, consider my comment bellow.

app/src/androidTest/java/com/github/brugapp/brug/ChatActivityPermissionDeniedTest.kt
DeepGreen1: I took this part form code implemented in Barrista. I think this is essentially due to versioning (we need to have at least the Marshmallow version for it to work), so, yes.

PR #316, opened 2022-05-19
tests + accessibiltiy 

ha2san:


ha2san:
In this PR some tests were done for the classes: ActivityResultModule and ProfileSettingsFragment. I still couldn't test all path but it is better than before. I also modified the addItem Activity where the layout was not the same on multiple device. It is supposed to be fixed now.

omarelmalki (approved):
Great job on making the app compatible across devices. Tested it on Nexus 5, Pixel 5 and Pixel C and everything seems to be accessible.

app/src/main/res/layout/activity_add_item.xml
```diff
@@ -1,169 +1,177 @@
...
+            android:id="@+id/add_item_button"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="@string/addButton"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintHorizontal_bias="0.439"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintVertical_bias="0.803" />
```
Mouniraki, line 173: By pulling the code and checking how it looks, it seems that the button is a bit off-centered in larger screens, is it possible to make it more centered ?

app/src/main/res/layout/activity_add_item.xml
```diff
@@ -1,169 +1,177 @@
...
+                android:layout_height="16dp"
+                android:text="@string/required"
+                android:textColor="@android:color/holo_red_dark"
+                android:textSize="12sp"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintHorizontal_bias="0.453"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:layout_constraintVertical_bias="0.226" />
```
Mouniraki, line 63: For the little "required" texts below the fields, I noticed that they are not aligned with the text fields in some screen sizes, is it possible to align them ? Also, I wonder if it is better to keep them below the fields or to put them above the fields, what do you think ?

Mouniraki (approved):
Great job ! Please consider my suggestions for the placement of some elements in the UI before merging, I think it would greatly improve the usability of the app.

app/src/main/res/layout/activity_add_item.xml
ha2san: yes of course

app/src/main/res/layout/activity_add_item.xml
ha2san: I think I will put them above the field

PR #322, opened 2022-05-20
Patch/nfclinks

pmcintyr:


app/src/main/java/com/github/brugapp/brug/ui/NFCScannerActivity.kt
```diff
@@ -87,31 +94,37 @@ open class NFCScannerActivity: AppCompatActivity() {
...
+        liveData(Dispatchers.IO){ emit(qrviewModel.parseTextAndCreateConv(editable, newcontext, firebaseAuth, firestore, firebaseStorage))}.observe(newcontext){ successState ->
             if(successState){
                 Toast.makeText(context, "Thank you ! The user will be notified.", Toast.LENGTH_LONG).show()
                 val myIntent = if(firebaseAuth.currentUser == null) Intent(this, SignInActivity::class.java) else Intent(this, ChatMenuActivity::class.java)
                 startActivity(myIntent)
-            } else Toast.makeText(context, "ERROR: An error has occurred, try again.", Toast.LENGTH_LONG).show() }
+                retval = true
+            } else {
+                if(!nfcContents.text.contains(firebaseAuth.currentUser!!.uid)){
+                    startActivity(Intent(this, ChatMenuActivity::class.java))
```
DeepGreen1, line 111: In the case where the NFC tag has been rewritten (eg: you write "Hey" instead of a valid User:Item string), it should be nice to have a toast displayed instead.

DeepGreen1 (approved):
Nice work on the NFC feature! I'll let you read the comment in the code. We can read a NFC from everywhere inside of the app (and not only on the NFC activity) which is really convenient. However, I noticed that we have to scan the NFC tag multiple times for the conversation to be created; I'll let you check if you can do something about it.
Also, we have to keep in mind to remove the button from the home menu once we are done implementing the NFC feature.

You're still running Cirrus but as it had passed the tests 2 commits ago, I'll trust you to merge once everything works. Again, nice job!

app/src/main/java/com/github/brugapp/brug/ui/NFCScannerActivity.kt
pmcintyr: Nice catch! I added both change requests in the [next commit](https://github.com/BrugApp/brug/pull/322/commits/5c7fe0d6fc13d2632d19b87552363c8863742eb9).

omarelmalki (changes requested):
Hi, I tested the feature on my phone and I was able to spot some issues:

Whenever I scan an NFC tag either from inside or outside the app, it opens the NFCScanner activity even if you just scanned the item.

When I scan my NFC tag from within the app when signed in without selecting any activity: 
It opens a page asking my to activate the tag but as soon as it text gets detected, it displays a notification on the same device and moves to chat menu activity without creating any conversation. If I spam the activate button before that happens, I manage to create a new item but it sometimes creates multiple items all with the same name and attributes without giving me the option to create a new item manually or link the NFC tag to an existing item.

When I scan the NFC when not signed in and activate it:
It sends me a toast telling me that a new item was created but how can you create an item that is not signed in. But it appears to have a correct behaviour when the tag contains a valid id (creates a conversation and sends location).

pmcintyr:
> Hi, I tested the feature on my phone and I was able to spot some issues:
> 
> Whenever I scan an NFC tag either from inside or outside the app, it opens the NFCScanner activity even if you just scanned the item.
> 
> When I scan my NFC tag from within the app when signed in without selecting any activity: It opens a page asking my to activate the tag but as soon as it text gets detected, it displays a notification on the same device and moves to chat menu activity without creating any conversation. If I spam the activate button before that happens, I manage to create a new item but it sometimes creates multiple items all with the same name and attributes without giving me the option to create a new item manually or link the NFC tag to an existing item.
> 
> When I scan the NFC when not signed in and activate it: It sends me a toast telling me that a new item was created but how can you create an item that is not signed in. But it appears to have a correct behaviour when the tag contains a valid id (creates a conversation and sends location).



> Hi, I tested the feature on my phone and I was able to spot some issues:
> 
> Whenever I scan an NFC tag either from inside or outside the app, it opens the NFCScanner activity even if you just scanned the item.
> 
> When I scan my NFC tag from within the app when signed in without selecting any activity: It opens a page asking my to activate the tag but as soon as it text gets detected, it displays a notification on the same device and moves to chat menu activity without creating any conversation. If I spam the activate button before that happens, I manage to create a new item but it sometimes creates multiple items all with the same name and attributes without giving me the option to create a new item manually or link the NFC tag to an existing item.
> 
> When I scan the NFC when not signed in and activate it: It sends me a toast telling me that a new item was created but how can you create an item that is not signed in. But it appears to have a correct behaviour when the tag contains a valid id (creates a conversation and sends location).

Hey, here are my answers regarding the issues you spotted:

_"Whenever I scan an NFC tag either from inside or outside the app, it opens the NFCScanner activity even if you just scanned the item."_ -> As we discussed in private in the code review follow-up, NFC scanning happens automatically (when enabled) when a tag is close enough to the device. This is the expected behavior and this means that opening the scanner is not contextual on previous scan actions unless they change the tag's text. 

_"When I scan my NFC tag from within the app when signed in without selecting any activity: It opens a page asking my to activate the tag but as soon as it text gets detected, it displays a notification on the same device and moves to chat menu activity without creating any conversation. If I spam the activate button before that happens, I manage to create a new item but it sometimes creates multiple items all with the same name and attributes without giving me the option to create a new item manually or link the NFC tag to an existing item."_ -> The creation of default item attributes seemed to be a problem, I falsely assumed we could modify item attributes from the items menu page but in my latest commit this has been corrected by adding an nfc button on the "add item" page. The self-notifications have also been removed.

_"When I scan the NFC when not signed in and activate it: It sends me a toast telling me that a new item was created but how can you create an item that is not signed in. But it appears to have a correct behaviour when the tag contains a valid id (creates a conversation and sends location)."_ -> This is also an issue, I have corrected this in my latest commit.

omarelmalki (changes requested):
Hi, great job on the changes. I can now indeed create an item with NFC.
However, when I can an item with a valid tuple of uid:iid with an anonymous user (I found an NFC tag), there are three main issues:
- First, it does not update the location of the item like scanning with a QR does.
- Second, it does not create a conversation for the owner of the item with the location sent to him like we have with the QR code
- Third, when we return from the NFC Scan activity with an anonymous user (not signed in), it opens the items menu as if we were signed in, when it should have simply returned to the SIgnInActivity.

pmcintyr:
> Hi, great job on the changes. I can now indeed create an item with NFC. However, when I can an item with a valid tuple of uid:iid with an anonymous user (I found an NFC tag), there are three main issues:
> 
> * First, it does not update the location of the item like scanning with a QR does.
> * Second, it does not create a conversation for the owner of the item with the location sent to him like we have with the QR code
> * Third, when we return from the NFC Scan activity with an anonymous user (not signed in), it opens the items menu as if we were signed in, when it should have simply returned to the SIgnInActivity.

Hey, here is what I did in reponse to these issues:

1) Mounir changed the way we set item location while I was already working on NFC, so I spent Tuesday updating the NFC classes correspondingly.

2) It does create a conversation, it just takes 10-15 seconds to appear. Regarding the location updates, that has been fixed in my latest commit.

3) This is an actual bug that I corrected manually in the code in my latest commit.

Thanks for being very observant for this PR, with the deadline approaching, I think that now it is time to merge.

omarelmalki (changes requested):
Hi, unfortunately the conversation created is still an Empty Conversation with no location or message sent and scanning the NFC tag still does not update the location of the object on my end.

PaulMcIntyre (approved):
All the changes have now been resolved.

omarelmalki (approved):
Good job ! It appears that an empty conversation was created between the demo account and the anonymous (guest) user (probably during some tests with buggy code) which apparently blocked the feature.
It works now. after deleting it Please try to improve the test coverage if possible.

PR #332, opened 2022-05-23
Revert "Complete cache implementation"

omarelmalki:
Reverts BrugApp/brug#308 while we find a solution to the low coverage issue.

EDIT: A fix to the issue was found in PR #334. There seems to be no need to revert it for now.


ha2san (approved):
I think it is really important that we find where the issue comes from. Until then I also think that a revert is needed

Mouniraki (approved):
Thank you for reacting quickly on this one, since before merging this part I had 80% coverage and didn't get any lower in the part I merged. I will investigate on why Cirrus gives us sometimes a very low coverage and come back to you to discuss about it.

PR #334, opened 2022-05-23
Bugfix attempt

omarelmalki:


ha2san (approved):
Nice bug fix, So according to your changes we have to always use the hilt module

Mouniraki (approved):
Nice job on this one, but now we must really understand why the checks sometimes fail silently without giving back any errors, so that it doesn't reproduce itself. I'm currently investigating on how Cirrus performs the checks from the .cirrus.yml script and also from the Gradle configuration files.

PR #342, opened 2022-05-25
Feature/UI design

hamzahass:
- Modified AddItemActivity's UI
- Removed search buttons on Top Bars
- Fixed home menu buttons
- Added permission tests for audio in chat


ha2san (changes requested):
Really nice work.
I just have some concern. The add item activity is a bit strange. For example, why the cursor is white ? And we can see a little text "description" in white in the description box. And can you make that the box for item name and type is not surrounded by blue like the description box please.

omarelmalki (approved):
Good job

ha2san (approved):
Thank you for the changes

PR #369, opened 2022-06-10
Clean-up comments in tests & remove bad assertions

Mouniraki:
This PR aims at removing all the comments + dead code present in the test classes, and removing useless assertions in test cases.

Some might still be present, feel free to check for yourself by pulling the code of the branch.

### DIFF
--- a/app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
+++ b/app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
@@ -1,17 +1,11 @@
 package com.github.brugapp.brug
 
-import android.app.Activity
-import android.content.Context
 import android.content.Intent
 import android.view.View
-import android.widget.EditText
 import android.widget.ImageView
-import android.widget.Spinner
-import android.widget.TextView
 import androidx.annotation.DrawableRes
 import androidx.core.graphics.drawable.toBitmap
 import androidx.recyclerview.widget.RecyclerView
-import androidx.test.annotation.UiThreadTest
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.Espresso.onData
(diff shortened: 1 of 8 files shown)


omarelmalki (approved):
Great cleanup work.

app/src/androidTest/java/com/github/brugapp/brug/AddItemTest.kt
```diff
@@ -111,6 +116,9 @@ class AddItemTest {
     fun cleanUp() {
         Intents.release()
         wipeItemsAndSignOut()
+        if(firebaseAuth.currentUser != null){
```
pmcintyr, line 119: I think this is a good idea, so that the other tests will not be influenced by the firebaseAuth state!

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
```diff
@@ -156,7 +156,7 @@ class ConvRepositoryTest {
                 2022, Month.MARCH, 23, 15, 30
             )
         )
-        val message = Message("BigBoy",dummyDate,"Wagwan")//why the test fails if I use this?
+        val message = Message("BigBoy",dummyDate,"Wagwan")
```
pmcintyr, line 159: Does this test no longer fail? I saw the previous comment is now removed.

app/src/androidTest/java/com/github/brugapp/brug/MessageRepositoryTest.kt
```diff
@@ -250,12 +245,6 @@ class MessageRepositoryTest {
         delay(2000)
 
         assertThat(BrugDataCache.getCachedMessageList("${USER_ID1}${USER_ID2}").value.isNullOrEmpty(), IsEqual(false))
```
pmcintyr, line 247: good job on removing the bulk comments that were still polluting our classes!

pmcintyr (approved):

app/src/androidTest/java/com/github/brugapp/brug/ConvRepositoryTest.kt
Mouniraki: It seemed that the test was now valid, it didn't fail when I ran it locally or on Cirrus.

